Issue #88, opened 2014-12-01
rebase to v0.5+ upstream


jdef:
Much has changed since 0.4x. A short list of **some** of the impacting changes appears below:

* services v2 (portals) support. aka ip-per-service model
* namespaces
* authentication and authorization framework (appears to be incubating)
* initial ipv6 support
* pod update support
* events

[[EDIT]]
* Release notes for [0.5x](https://github.com/GoogleCloudPlatform/kubernetes/releases/tag/v0.5)
* Release notes for [0.6x](https://github.com/GoogleCloudPlatform/kubernetes/releases/tag/v0.6.1)

jdef:
resolved with merge of #102

Issue #59, opened 2014-11-05
services should not require pods and pod-templates to declare explicit hostPort's


jdef:
since merging #54 services can now connect to pods but only if those pods (or pod templates) are declared with a hostPort != 0. not including a hostPort definition is the same as assigning a value of zero, which triggers the default k8s behavior of not publishing the containerPort to the host at all. specifying a hostPort allows a service to become attached to the hostIP:hostPort, but this (a) requires that users are explicitly maintaining a list of available/consumed host ports, and; (b) limits the number of instances of said pod/pod-template to 1 per-host.

the enhancement here is to somehow allow an end user to specify that their pod/pod-template wants a hostPort, but without explicitly naming that port. mesos supports this by offering up a range of available ports to the k8s-mesos scheduler, but there's no way for the scheduler to know when a user wants a hostPort dynamically assigned from the available port range.

i've previously suggested a magic hostPort of 1 to indicate to the scheduler that dynamic hostPort assignment is desired by the end-user. upon observing a value of 1 the scheduler would dynamically allocate a host port from the available offers and update the pod's desired state to reflect the newly allocated port.

this seems extremely hackish, and limits the real-world use of hostPort 1 by this particular mesos scheduler. however it is backwards compatible with k8s existing use of hostPort 0 as well as with the existing JSON API (no new fields are required).

an alternative approach is to require an end-user to apply a magic label (something like "dynamic-host-port" with, perhaps, a boolean value of true/false) to obtain this behavior. initially this seems cleaner than the magic hostPort=1 approach, but seems to start us down a path of using labels to obtain specific scheduler behavior -- and I'm not sure that's the direction we want to go.

@adam-mesos @ConnorDoyle Thoughts?

jdef:
since the kubernetes architecture has changed, and the kubelet is now responsible for pod rectification, it seems MUCH safer to allocate random host ports for each container port so that services may be redirected to them. That means reusing lots of the code from the expose_services_label branch that deals with hostport allocation when hostport == 0.

As part of this, thinking about an optional label (k8sm.mesosphere.io/service-ports ?) that if specified, restricts which dynamic host ports (hostport == 0) are actually allocated from an offer (as opposed to the default, which would be all of them). So if a container declared:

`ports: [{ containerPort: 80}, {containerPort: 443}, {containerPort:9999}]`

..  then it could also specify the label:

`labels=[{name:"k8sm.mesosphere.io/service-ports", value:"80,443"}]`

[...]

PR #103, opened 2014-12-08
k8sm/expose label to indicate that hostPort==0 should be exposed on an available port

jdef:
This branch attempts to implement a solution for #59.

A special label `k8sm:expose` may be used to auto-publish a container port to the host, using any non-conflicting port in a mesos resource offer. Non-conflicting means that (a) it's in the "ports" resource range of the offer, and; (b) no other container in the pod has directly specified that pod as its HostPort.

`k8sm:expose` currently accepts a single value `'*'`, which auto-publishes all container ports with HostPort == 0 (wildcard ports). Current thinking is that the value of this label could also be a comma-delimited list of Port.ContainerPort's (int or string) to which the auto-publish rule would be applied.

TODOs:
- [ ] endpoint controller currently uses DesiredState to derive endpoints: this is a **problem** because currently the pod_task implementation (purposefully) doesn't overwrite DesiredState, which means that findPort() will return 0 for wildcard ports -- not good. perhaps our implementation of endpoint controller could query /boundPods on the kubelet? (would need to wait for rebase to 05x (see #88)). **Alternative:** "filter" the GetPods and ListPods results, updating the DesiredState and CurrentState for all returned objects such that wildcard host port overrides are updated on the fly. This makes the endpoint controller happy, and allows us to keep a "cleanroom" DesiredState spec stored with the task.
- [x] consider `k8sm/expose` as an alternative label? there seems to be a tendency in k8s for using `'/'` as a delimiter for other things.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,8 @@ WITH_MESOS_CGO_FLAGS :=  \
 	  CGO_CXXFLAGS="$(CGO_CXXFLAGS)" \
 	  CGO_LDFLAGS="$(CGO_LDFLAGS)"
 
+WITH_MESOS_LD_LIBRARY_PATH := LD_LIBRARY_PATH=$(WITH_MESOS_DIR)/lib
+
 endif
 
 export SHELL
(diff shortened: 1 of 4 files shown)


jdef:
This approach may become less necessary as support for ipv6 will probably launch with docker 1.5, and kubernetes is probably not very far behind.

jdef:
closing this out for now, will revisit later. many changes from this PR have already been merged, except those that actually change the allocation strategy. recommendation is to start a new PR at some later point

PR #112, opened 2015-01-06
Refactor pod registry

jdef:
**WARNING:** major refactoring ahead. This branch removes the customized API server in favor of the stock implementation. As such, there are major consequences to scheduling and minion registration/tracking. Tutorials will need updating (see #122).

TODOs:
- [x] refactor scheduler
- [x] refactor mesos cloud provider implementation
- [x] update README / walkthrough
- [x] single pod deploy working
- [x] restore scheduling error handler backoff
- [x] avoid re-queuing older pod state over the top of newer pod state
- [x] guestbook example working
- [x] unit testing (started, see #123)
- [x] burn-in testing


jdef:
When apply binding fails, there's a 3-way race between:
- the k8s scheduling loop
- the async scheduling error handler func
- the pod FIFO reflector

Apply binding checks bound pod constraints, AFTER the pod.Host is updated to a non-empty string. If there's a problem with the constraint check, then the pod.Host is reset to "" (there are now two write/updates sent to the pod FIFO reflector). An error is then propagated back to the scheduler, which triggers the async scheduling error handler. However, since the pod.Host was reset to "", the FIFO reflector will have added it to the schedulable-pods queue causing the k8s scheduling loop to pick it up immediately and attempt to schedule it. Rinse and repeat to generate several simultaneous async error handlers running at the same time as the main scheduling loop scheduler func -- all for the same pod.

The eager back-off breakout approach taken by k8s-mesos aggravates this further, though the root of the problem is actually the upstream design. There are two paths by which pods are re-added to the scheduler queue (the FIFO reflector, and the error handler) and the backoff feature of the error handler is completely bypassed when the FIFO reflector picks up the updated pod metadata in this particular failure scenario.

I think a partial solution may be present in the form of a delay-oriented FIFO, where the delay is proportional to the "heat" of a pod. Heat would correlate with the number of scheduling attempts per unit time for a given pod. More thought needed here.

jdef:
Following up from my last comment: the scheduling race dilemma was handled in the following way:
* use separate pod store for current pod state vs. pods queued for scheduling
* scheduling errors are handled synchronously: backoff waits are performed by a priority queue
* an async goroutine publishes unscheduled pods (from the pod store) to the scheduling queue
* the Schedule, Bind, and Error implementations are more aggressive in checking for valid pod status

cmd/k8sm-scheduler/main.go
```diff
@@ -0,0 +1,292 @@
...
+	return
+}
+
+func getUsername() (username string, err error) {
+	username = *mesosUser
+	if username == "" {
+		if u, err := user.Current(); err == nil {
+			username = u.Username
+			if username == "" {
+				username = "root"
```
jdef: extract constant: "root"

pkg/cloud/mesos/client.go
```diff
@@ -0,0 +1,164 @@
...
+			} `json:"slaves"`
+		}
+		state := &State{}
+		err = json.Unmarshal(blob, state)
+		if err != nil {
+			return err
+		}
+		for _, slave := range state.Slaves {
+			if slave.Pid != "" {
+				if parts := strings.SplitN(slave.Pid, "@", 2); len(parts) == 2 && len(parts[1]) > 0 {
```
jdef: TODO(jdef) use upid.UPID (from pure bindings, once we migrate) to parse this

pkg/cloud/mesos/client.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	//TODO(jdef) should parallelize this
+	results := []string{}
+	for _, slave := range slaves {
+		if found, err := c.slaveRunningKubeletExecutor(ctx, slave); found {
+			// parse the host from the slave host:port
+			if parts := strings.SplitN(slave, ":", 2); len(parts) > 0 && len(parts[0]) > 0 {
```
jdef: use golang net.SplitHostPort

Issue #124, opened 2015-01-21
rebase to v0.8x+ upstream


jdef:
TODOs
- [x] cmd/kubelet/kubelet.go :: largely refactored (most of it went into config+standalone)
- [x] plugin/cmd/kube-scheduler/scheduler.go :: minor changes
- [x] pkg/service/endpoints_controller.go :: minor changes
- [x] plugin/pkg/scheduler/factory/factory.go :: investigate plugins, auth, alg providers (#121); will not address as part of this rebase
- [x] single pod
- [x] guestbook
- [x] burn-in testing

jdef:
closing out since #132 merged

PR #132, opened 2015-01-26
Rebase to k8s 0.8x

jdef:
progress tracked in #124


jdef:
Ready for merge, PTAL @ConnorDoyle @vladimirvivien

cmd/k8sm-executor/main.go
```diff
@@ -54,135 +48,120 @@ var (
 	minimumGCAge            = flag.Duration("minimum_container_ttl_duration", 0, "Minimum age for a finished container before it is garbage collected.  Examples: '300ms', '10s' or '2h45m'")
 	maxContainerCount       = flag.Int("maximum_dead_containers_per_container", 5, "Maximum number of old instances of a container to retain per container.  Each container takes up some disk space.  Default: 5.")
 	authPath                = flag.String("auth_path", "", "Path to .kubernetes_auth file, specifying how to authenticate to API server.")
+	cAdvisorPort            = flag.Uint("cadvisor_port", 4194, "The port of the localhost cAdvisor endpoint")
```
adam-mesos: Have you tested cAdvisor with k8sm? Just curious.

cmd/k8sm-executor/main.go
```diff
@@ -54,135 +48,120 @@ var (
...
+		RootDirectory:           *rootDirectory,
+		NetworkContainerImage:   *networkContainerImage,
+		SyncFrequency:           *syncFrequency,
+		RegistryPullQPS:         *registryPullQPS,
+		RegistryBurst:           *registryBurst,
+		MinimumGCAge:            *minimumGCAge,
+		MaxContainerCount:       *maxContainerCount,
+		ClusterDomain:           *clusterDomain,
+		ClusterDNS:              clusterDNS,
+		Port:                    *port, // TODO(nnielsen): Don't hardwire port, but use port from resource offer.
```
adam-mesos: This doesn't look hardwired. What does this TODO mean?

cmd/k8sm-executor/main.go
```diff
@@ -54,135 +48,120 @@ var (
...
 	}
 
-	// Log the events locally too.
-	record.StartLogging(log.Infof)
+	driver := new(mesos.MesosExecutorDriver)
 
-	capabilities.Initialize(capabilities.Capabilities{
-		AllowPrivileged: *allowPrivileged,
-	})
+	//HACK(jdef) largely hacked from the k8s standalone package, createAndInitKubelet
```
adam-mesos: Did you mean to leave this //HACK comment in? Should it be replaced with an appropriate TODO or an @see reference?

cmd/k8sm-executor/main.go
```diff
@@ -54,135 +48,120 @@ var (
...
 
-	dockerClient, err := docker.NewClient(getDockerEndpoint())
-	if err != nil {
-		log.Fatal("Couldn't connnect to docker.")
-	}
+		log.V(2).Infof("Initialize executor driver...")
+		driver.Init()
 
-	hostname := getHostname()
+		k.BirthCry()
```
adam-mesos: Weird. Couldn't they have just called it Register or Announce?

hack/patches/k8s---issue128.patch
```diff
@@ -0,0 +1,141 @@
...
+ 	// start the kubelet
+ 	go util.Forever(func() { k.Run(cfg.Updates()) }, 0)
+ 
+ 	// start the kubelet server
+ 	if kc.EnableServer {
+ 		go util.Forever(func() {
+-			kubelet.ListenAndServeKubeletServer(k, net.IP(kc.Address), kc.Port, kc.EnableDebuggingHandlers)
+-		}, 0)
++			k.ListenAndServe(net.IP(kc.Address), kc.Port, kc.EnableDebuggingHandlers)
++		}, 10*time.Second)
```
adam-mesos: Maybe add the "We don't want 0s because if we can't get the port right away then we generate **tons** of logging and fill up the disk." comment here?

cmd/k8sm-executor/main.go
```diff
@@ -6,26 +6,20 @@ import (
...
 	"strings"
+	"sync"
 	"time"
 
-	"github.com/GoogleCloudPlatform/kubernetes/pkg/capabilities"
-	"github.com/GoogleCloudPlatform/kubernetes/pkg/client"
-	"github.com/GoogleCloudPlatform/kubernetes/pkg/client/record"
-	"github.com/GoogleCloudPlatform/kubernetes/pkg/clientauth"
-	"github.com/GoogleCloudPlatform/kubernetes/pkg/health"
 	_ "github.com/GoogleCloudPlatform/kubernetes/pkg/healthz"
```
vladimirvivien: Is healthz package here for package init() purpose or is it not being used?

pkg/executor/server.go
```diff
@@ -91,7 +92,8 @@ func NewServer(host HostInterface, enableDebuggingHandlers bool, ns string) Serv
 func (s *Server) InstallDefaultHandlers() {
 	healthz.InstallHandler(s.mux)
 	profile.InstallHandler(s.mux)
-	s.mux.HandleFunc("/podInfo", s.handlePodInfo)
+	s.mux.HandleFunc("/podInfo", s.handlePodInfoOld)
+	s.mux.HandleFunc("/api/v1beta1/podInfo", s.handlePodInfoVersioned)
```
adam-mesos: I thought we'd moved past v1beta1?

pkg/executor/server.go
```diff
@@ -299,23 +311,21 @@ func (s *Server) serveStats(w http.ResponseWriter, req *http.Request) {
...
-				Name: components[1],
-				// TODO(k8s): I am broken
+				Name:        components[1],
 				Namespace:   api.NamespaceDefault,
 				Annotations: map[string]string{kubelet.ConfigSourceAnnotationKey: s.sourcename},
 			},
 		})
 		stats, err = s.host.GetContainerInfo(podFullName, "", components[2], &query)
-	case 4:
+	case 5:
```
adam-mesos: Why did this become case 5, and where is case 4?

pkg/scheduler/plugin.go
```diff
@@ -157,7 +156,7 @@ func (b *binder) bind(ctx api.Context, binding *api.Binding, task *PodTask) (err
 
 	if err = b.prepareTaskForLaunch(ctx, binding.Host, task); err == nil {
 		log.V(2).Infof("Attempting to bind %v to %v", binding.PodID, binding.Host)
-		if err = b.client.Post().Namespace(api.Namespace(ctx)).Path("bindings").Body(binding).Do().Error(); err == nil {
+		if err = b.client.Post().Namespace(api.Namespace(ctx)).Resource("bindings").Body(binding).Do().Error(); err == nil {
```
vladimirvivien: Break the chained accessors from b.client up a bit, no? this would help with debug if something gets broken.

jdef: I see your point. This line is basically copy/pasted from k8s code. Upon further inspection, k8s implements a nice error propagation pattern in Request objects:

https://github.com/GoogleCloudPlatform/kubernetes/blob/release-0.8/pkg/client/request.go

I'm probably going to leave this as-is.

pkg/executor/server.go
jdef: k8s code, there is no spoon, I mean case 4. Valid URL paths handled by this target basically have 1, 3, or 5 path segments. Anything else is not supported.

pkg/executor/server.go
jdef: k8s code: I don't know why they use v1beta1 instead of v1beta2, but last I checked those are the only two "publishable" types that clients can use for interacting with the API.

cmd/k8sm-executor/main.go
jdef: init()

hack/patches/k8s---issue128.patch
jdef: Good point, I'll include that when I upstream this patch.

cmd/k8sm-executor/main.go
jdef: agreed

cmd/k8sm-executor/main.go
jdef: the point was that the port should come from the mesos `ports` resource range, and not from the command line. i'll update the wording since it's odd now that the code has shifted.

cmd/k8sm-executor/main.go
jdef: Nope, but since the executor doesn't spin it up (like the proxy) we can't depend on it being there.

cmd/k8sm-executor/main.go
jdef: Perhaps it reflects their excitement and optimism in bringing a new kubelet into the world.

PR #142, opened 2015-02-04
Pure bindings

jdef:
TODOs:
- [x] compiles and runs replication controller
- [ ] test authentication support
- [ ] test guestbook example
- [ ] burn-in testing

### DIFF
--- a/cmd/k8sm-executor/main.go
+++ b/cmd/k8sm-executor/main.go
@@ -22,7 +22,7 @@ import (
 	"github.com/GoogleCloudPlatform/kubernetes/pkg/version/verflag"
 	"github.com/fsouza/go-dockerclient"
 	log "github.com/golang/glog"
-	"github.com/mesos/mesos-go/mesos"
+	edriver "github.com/mesos/mesos-go/executor"
 	"github.com/mesosphere/kubernetes-mesos/pkg/executor"
 )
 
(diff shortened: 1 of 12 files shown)


cmd/k8sm-executor/main.go
```diff
@@ -154,18 +153,15 @@ func main() {
 	standalone.RunKubelet(&kcfg, builder)
 
 	log.Infoln("Waiting for driver initialization to complete")
-	<-initialized
-	defer driver.Destroy()
 
 	// block until driver is shut down
-	driver.Join()
+	select {}
```
vladimirvivien: Interesting trick.  Will that cause the driver to block instead of driver.Join() ?

pkg/scheduler/pod_task.go
```diff
@@ -55,8 +56,8 @@ func (t *PodTask) FillFromDetails(details *mesos.Offer) error {
 	t.TaskInfo.TaskId = newTaskID(t.ID)
 	t.TaskInfo.SlaveId = details.GetSlaveId()
 	t.TaskInfo.Resources = []*mesos.Resource{
-		mesos.ScalarResource("cpus", containerCpus),
-		mesos.ScalarResource("mem", containerMem),
+		mutil.NewScalarResource("cpus", containerCpus),
```
vladimirvivien: Cool, you are using the mesos util package.

cmd/k8sm-executor/main.go
vladimirvivien: I did a search for "empty select" got my answer.

PR #150, opened 2015-02-15
Forcing tag change since it is required now

bobrik:
Since 1.4 (docker/docker#8511):

```
FATA[0000] Error response from daemon: Conflict: Tag build-base is already set to image 25deb5eeb2834533442d3a136ef2e150610596926b21c0b45597fe454f854dd0, if you want to replace it, please use -f option
```

Happens when you run `hack/dockerbuild/build`.


jdef:
thanks for the PR

PR #163, opened 2015-03-03
Better support for task reconciliation

### LINKED ISSUE #73: implement task reconciliation
https://github.com/apache/mesos/blob/master/docs/reconciliation.md

### LINKED ISSUE #162: Persist Bindings annotations alongside pod Host assignment
PR submitted upstream for review

jdef:
- fixes #162 
- resolves #73

### DIFF
--- a/pkg/scheduler/podtask/pod_task.go
+++ b/pkg/scheduler/podtask/pod_task.go
@@ -206,9 +206,9 @@ func RecoverFrom(pod *api.Pod) (*T, bool, error) {
 		return nil, false, fmt.Errorf("illegal argument: pod was nil")
 	}
 	if pod.Status.Host == "" || len(pod.Annotations) == 0 {
-		//TODO(jdef) if Status.Host != "" then it's likely that the task has launched
-		//but is not yet bound -- so annotations may be on the way. The binding may also
-		//have failed but we haven't been processed the TASK_FAILED yet.
+		// if Status.Host != "" then it's likely that the task has launched
+		// but is not yet bound -- so annotations may be on the way. The binding
+		// may also have failed but we haven't been processed the TASK_FAILED yet.
 		log.V(1).Infof("skipping recovery for unbound pod %v/%v", pod.Namespace, pod.Name)
 		return nil, false, nil
 	}
--- a/pkg/scheduler/podtask/pod_task.go
+++ b/pkg/scheduler/podtask/pod_task.go
@@ -10,6 +10,7 @@ import (
 	mesos "github.com/mesos/mesos-go/mesosproto"
 	mutil "github.com/mesos/mesos-go/mesosutil"
 	"github.com/mesosphere/kubernetes-mesos/pkg/offers"
+	annotation "github.com/mesosphere/kubernetes-mesos/pkg/scheduler/meta"
 	"github.com/mesosphere/kubernetes-mesos/pkg/scheduler/metrics"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/scheduler/podtask/pod_task.go
```diff
@@ -183,6 +184,98 @@ func New(ctx api.Context, pod *api.Pod, executor *mesos.ExecutorInfo) (*T, error
...
+//
+// at this time task reconstruction is only supported for pods that have been annotated with
+// binding metadata, which implies that they've previously been associated with a task and
+// that mesos knows about it.
+//
+func RecoverFrom(pod *api.Pod) (*T, bool, error) {
+	if pod == nil {
+		return nil, false, fmt.Errorf("illegal argument: pod was nil")
+	}
+	if pod.Status.Host == "" || len(pod.Annotations) == 0 {
```
jdef: If the binding fails, then Host == "" but pod status == Failed, so we can actually check for that here.

pkg/scheduler/scheduler.go
```diff
@@ -223,7 +226,11 @@ func (k *KubernetesScheduler) StatusUpdate(driver bindings.SchedulerDriver, task
 		}
 		fallthrough
 	case mesos.TaskState_TASK_KILLED:
-		k.taskRegistry.UpdateStatus(taskStatus)
+		_, state := k.taskRegistry.UpdateStatus(taskStatus)
+		//TODO(jdef) what if I receive this after a TASK_LOST or TASK_KILLED? I don't want to reincarnate then..
```
jdef: It's possible that master may report TASK_LOST and subsequently report a non-terminal status. TASK_LOST is special this way, unlike the other terminal statuses. We already track *finished* tasks (for some reason) -- we could just as easily track *lost* tasks and handle this TASK_LOST-followed-by-TASK_STARTING more gracefully.

pkg/scheduler/scheduler.go
```diff
@@ -248,6 +255,53 @@ func (k *KubernetesScheduler) StatusUpdate(driver bindings.SchedulerDriver, task
 	}
 }
 
+func (k *KubernetesScheduler) handleUnknownStartedTask(driver bindings.SchedulerDriver, taskStatus *mesos.TaskStatus) {
+	//TODO(jdef) attempt to recover task from pod info:
```
jdef: This isn't really a TODO any more, but it does need unit testing

pkg/scheduler/scheduler.go
```diff
@@ -350,3 +404,25 @@ func (r *Reconciler) Run(driver bindings.SchedulerDriver) <-chan struct{} {
 	}
 	return nil
 }
+
+func (ks *KubernetesScheduler) recoverTasks() error {
```
jdef: Needs unit testing

pkg/scheduler/service/service.go
```diff
@@ -300,9 +302,11 @@ func (s *SchedulerServer) Run(_ []string) error {
 		log.Fatalf("Failed to start driver: %v", err)
 	}
 
-	//TODO(jdef) we need real task reconciliation at some point
-	log.V(1).Info("Clearing old pods from the registry")
-	clearOldPods(client)
+	/*
```
jdef: Should we enable this code with a flag (e.g. `--clean_slate`)? If so, it would have to be invoked before mesosPodScheduler.Init() which is recovering tasks from the pod state.

pkg/scheduler/podtask/pod_task.go
jdef: Also, should the scheduler check for non-terminal pod.Status.Phase?

PR #178, opened 2015-03-15
Task status metrics

### LINKED ISSUE #174: add metrics related to reconciliation
- count of requested explicit reconciliations
- count of requested implicit reconciliations
- count of executed implicit reconciliations
- count of completed explicit reconciliations
- count of cancelled explicit reconciliations
- latency for explicit reconciliation (may require using/tracking reconcileTime per-task, instead of updateTime)

jdef:
- fixes #174

### DIFF
--- a/pkg/scheduler/metrics/metrics.go
+++ b/pkg/scheduler/metrics/metrics.go
@@ -34,6 +34,37 @@ var (
 		},
 		[]string{"source", "reason", "state"},
 	)
+	ReconciliationLatency = prometheus.NewSummary(
+		prometheus.SummaryOpts{
+			Subsystem: schedulerSubsystem,
+			Name:      "reconciliation_latency_microseconds",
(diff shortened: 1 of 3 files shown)


jdef:
I've observed metrics like this:
```
scheduler_status_updates{reason="none",source="none",state="TASK_KILLED"} 16
scheduler_status_updates{reason="none",source="none",state="TASK_RUNNING"} 16
scheduler_status_updates{reason="none",source="none",state="TASK_STARTING"} 16
```

Suspect it's because the go executor driver bindings are not enforcing `SOURCE_EXECUTOR` in `TaskStatus.Source`. But the native bindings don't do that yet either:

https://github.com/apache/mesos/blob/master/src/exec/exec.cpp#L490

jdef:
The missing TaskStatus.Source bug has been filed as https://issues.apache.org/jira/browse/MESOS-2499

Issue #114, opened 2015-01-06
integrate naive mesos cloud provider into k8s


jdef:

jdef:
:tada: we're in!

tsenart:
:8ball:

ConnorDoyle:
:cocktail:

Issue #175, opened 2015-03-13
mesos cloud provider should be configurable via config file, not pflags


jdef:
k8s cloud providers are configured via files. the mesos provider uses `mesos_master`

jdef:
resolved by #210

PR #210, opened 2015-04-08
Upstream cloud provider prep

ConnorDoyle:
Addresses a pre-requisite for moving the cloud provider implementation upstream: reading config from a file.  Related to #114.

- [X] Switch from `--mesos-master` flag to read this value from a config file
- [x] Test coverage
- [x] Documentation updates

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,15 +85,6 @@ If you are running an older version of etcd you may want to consider an upgrade.
 #### kubernetes-mesos
 
 Ensure that your Mesos cluster is started.
-If you're running a standalone Mesos master on `${servicehost}` then set:
-```shell
-$ export mesos_master=${servicehost}:5050
-```
-
-Or if you have multiple Mesos masters registered with a Zookeeper cluster then set:
-```shell
-$ export mesos_master=zk://${zkserver1}:2181,${zkserver2}:2181,${zkserver3}:2181/mesos
-```
 
 Create a configuration file for the Mesos cloud provider:
 
@@ -105,7 +96,7 @@ The file content should have the following format:
 
 ```ini
 [mesos-cloud]
-	mesos-master        = leader.mesos:5050
+	mesos-master        = host:port
 	http-client-timeout = 5s
 	state-cache-ttl     = 20s
 ```
@@ -125,7 +116,6 @@ Fire up the kubernetes-mesos framework components (yes, these are **all** requir
 ```shell
 $ ./bin/km apiserver \
   --address=${servicehost} \
-  --mesos_master=${mesos_master} \
   --etcd_servers=http://${servicehost}:4001 \
   --portal_net=10.10.10.0/24 \
   --port=8888 \
@@ -134,11 +124,10 @@ $ ./bin/km apiserver \
 
 $ ./bin/km controller-manager \
   --master=$servicehost:8888 \
-  --mesos_master=${mesos_master}
+ --cloud_config=./mesos-cloud.conf
 
 $ ./bin/km scheduler \
   --address=${servicehost} \
-  --mesos_master=${mesos_master} \
   --etcd_servers=http://${servicehost}:4001 \
   --mesos_user=root \
   --api_servers=$servicehost:8888
--- a/README.md
+++ b/README.md
@@ -95,6 +95,31 @@ Or if you have multiple Mesos masters registered with a Zookeeper cluster then s
 $ export mesos_master=zk://${zkserver1}:2181,${zkserver2}:2181,${zkserver3}:2181/mesos
 ```
 
+Create a configuration file for the Mesos cloud provider:
+
+```shell
+$ edit ./mesos-cloud.conf
+```
+
+The file content should have the following format:
+
+```ini
+[mesos-cloud]
+	mesos-master        = leader.mesos:5050
+	http-client-timeout = 5s
+	state-cache-ttl     = 20s
+```
+
+**Mesos Cloud Provider Configuration Options:**
+
+- `mesos-master`: Location of the Mesos master.  This value can take the
+  form `host:port` or a canonical Zookeeper URL
+(`zk://host1:port1,host2:port2/your/znode/path`)
+- `http-client-timeout`: Time to wait for a response from the Mesos
+  master when querying the state.
+- `state-cache-ttl`: Maximum age of the cached Mesos state, after which
+  new values will be fetched.
+
 Fire up the kubernetes-mesos framework components (yes, these are **all** required for a working framework):
 
 ```shell
@@ -105,6 +130,7 @@ $ ./bin/km apiserver \
   --portal_net=10.10.10.0/24 \
   --port=8888 \
   --cloud_provider=mesos
+  --cloud_config=./mesos-cloud.conf
 
 $ ./bin/km controller-manager \
   --master=$servicehost:8888 \
--- a/README.md
+++ b/README.md
@@ -124,9 +124,10 @@ $ ./bin/km apiserver \
 
 $ ./bin/km controller-manager \
   --master=$servicehost:8888 \
- --cloud_config=./mesos-cloud.conf
+  --cloud_config=./mesos-cloud.conf
 
 $ ./bin/km scheduler \
+  --mesos_master=${mesos_masster} \
   --address=${servicehost} \
   --etcd_servers=http://${servicehost}:4001 \
   --mesos_user=root \
--- a/pkg/cloud/mesos/client.go
+++ b/pkg/cloud/mesos/client.go
@@ -19,8 +19,7 @@ import (
 )
 
 const (
-	mesosHttpClientTimeout = 10 * time.Second //TODO(jdef) configurable via fiag?
-	nodesCacheTTL          = 5 * time.Second  //TODO(jdef) configurable somehow?
+	defaultClusterName = "mesos"
 )
 
 var (
@@ -30,10 +29,10 @@ var (
 type mesosClient struct {
 	masterLock    sync.RWMutex
 	master        string // host:port formatted address
-	client        *http.Client
+	httpClient    *http.Client
 	tr            *http.Transport
 	initialMaster <-chan struct{} // signal chan, closes once an initial, non-nil master is found
-	nodes         *nodesCache
+	state         *stateCache
 }
 
 type slaveNode struct {
@@ -41,44 +40,80 @@ type slaveNode struct {
 	resources *api.NodeResources
 }
 
-type nodesCache struct {
-	sync.Mutex
-	expiresAt time.Time
-	cached    []*slaveNode
-	err       error
-	ttl       time.Duration
-	refill    func(context.Context) ([]*slaveNode, error)
+type mesosState struct {
+	clusterName string
+	nodes       []*slaveNode
 }
 
-// return the cached list of slave nodes. if needed, the cache is refreshed prior to returning the results.
-func (c *nodesCache) get(ctx context.Context) ([]*slaveNode, error) {
+type stateCache struct {
+	sync.Mutex
+	expiresAt time.Time
+	cached    *mesosState
+	err       error
+	ttl       time.Duration
+	refill    func(context.Context) (*mesosState, error)
+}
+
+// Reload the state cache if it has expired.
+func (c *stateCache) reloadCache(ctx context.Context) {
 	now := time.Now()
 	c.Lock()
 	defer c.Unlock()
 	if c.expiresAt.Before(now) {
+		log.V(4).Infof("Reloading cached Mesos state")
 		c.cached, c.err = c.refill(ctx)
 		c.expiresAt = now.Add(c.ttl)
 	} else {
-		log.V(4).Infof("returning cached slave list")
+		log.V(4).Infof("Using cached Mesos state")
 	}
+}
+
+// Returns the cached Mesos state.
+func (c *stateCache) getCachedState(ctx context.Context) (*mesosState, error) {
+	c.reloadCache(ctx)
 	return c.cached, c.err
 }
 
-func newMesosClient(md detector.Master) (*mesosClient, error) {
+// Returns the cached Mesos cluster name.
+func (c *stateCache) getClusterName(ctx context.Context) (string, error) {
+	cached, err := c.getCachedState(ctx)
+	return cached.clusterName, err
+}
+
+// Returns the cached list of slave nodes.
+func (c *stateCache) getNodes(ctx context.Context) ([]*slaveNode, error) {
+	cached, err := c.getCachedState(ctx)
+	return cached.nodes, err
+}
+
+func newMesosClient(
+	md detector.Master,
+	mesosHttpClientTimeout, stateCacheTTL time.Duration) (*mesosClient, error) {
+
 	tr := &http.Transport{}
+	httpClient := &http.Client{
+		Transport: tr,
+		Timeout:   mesosHttpClientTimeout,
+	}
+	return createMesosClient(md, httpClient, tr, stateCacheTTL)
+}
+
+func createMesosClient(
+	md detector.Master,
+	httpClient *http.Client,
+	tr *http.Transport,
+	stateCacheTTL time.Duration) (*mesosClient, error) {
+
 	initialMaster := make(chan struct{})
 	client := &mesosClient{
-		client: &http.Client{
-			Transport: tr,
-			Timeout:   mesosHttpClientTimeout,
-		},
+		httpClient:    httpClient,
 		tr:            tr,
 		initialMaster: initialMaster,
-		nodes: &nodesCache{
-			ttl: nodesCacheTTL,
+		state: &stateCache{
+			ttl: stateCacheTTL,
 		},
 	}
-	client.nodes.refill = client.pollMasterForSlaves
+	client.state.refill = client.pollMasterForState
 	first := true
 	if err := md.Detect(detector.OnMasterChanged(func(info *mesos.MasterInfo) {
 		client.masterLock.Lock()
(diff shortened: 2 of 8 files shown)


pkg/cloud/mesos/mesos.go
```diff
@@ -72,9 +89,30 @@ func (c *MesosCloud) Zones() (cloudprovider.Zones, bool) {
...
+
+// Master gets back the address (either DNS name or IP address) of the master node for the cluster.
+func (c *MesosCloud) Master(clusterName string) (string, error) {
+	clusters, err := c.ListClusters()
+	if err != nil {
+		return "", err
+	}
+	for _, name := range clusters {
+		if name == clusterName {
+			return c.client.master, nil
```
ConnorDoyle: Note to self, need to strip the port from this value.

jdef: `c.client.master` may be empty if there's no current leader

also, see : http://golang.org/pkg/net/#SplitHostPort

pkg/cloud/mesos/client.go
```diff
@@ -19,8 +19,9 @@ import (
 )
 
 const (
-	mesosHttpClientTimeout = 10 * time.Second //TODO(jdef) configurable via fiag?
-	nodesCacheTTL          = 5 * time.Second  //TODO(jdef) configurable somehow?
+	mesosHttpClientTimeout = 10 * time.Second //TODO(jdef) configurable via flag?
```
jdef: might be nice to get these into the config file, now that we have one :)

ConnorDoyle: +1, will do.

pkg/cloud/mesos/mesos.go
ConnorDoyle: Cool, good point.

jdef:
xref https://github.com/mesosphere/kubernetes-mesos/issues/175

pkg/cloud/mesos/client.go
```diff
@@ -41,30 +42,53 @@ type slaveNode struct {
...
+	return cached.clusterName, err
+}
+
+// Returns the cached list of slave nodes.
+func (c *stateCache) getNodes(ctx context.Context) ([]*slaveNode, error) {
+	cached, err := c.getCachedState(ctx)
+	return cached.nodes, err
+}
+
+func newMesosClient(md detector.Master, mesosHttpClientTimeout time.Duration, stateCacheTTL time.Duration) (*mesosClient, error) {
```
jdef: small nit: more idiomatic to declare:
```go
func newMesosClient(md detector.Master, mesosHttpClientTimeout, stateCacheTTL time.Duration) (*mesosClient, error) {
```

ConnorDoyle: Thanks James, still learning the ropes.  WIll fix.

pkg/cloud/mesos/mesos.go
```diff
@@ -52,29 +56,49 @@ func newMesosCloud() (*MesosCloud, error) {
...
+
+// Master gets back the address (either DNS name or IP address) of the master node for the cluster.
+func (c *MesosCloud) Master(clusterName string) (string, error) {
+	clusters, err := c.ListClusters()
+	if err != nil {
+		return "", err
+	}
+	for _, name := range clusters {
+		if name == clusterName {
+			return c.client.master, nil // TODO(CD): strip port here
```
ConnorDoyle: Note to self, still need to strip the port here.

pkg/cloud/mesos/mesos_test.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"bytes"
+	"fmt"
+	"testing"
+	"time"
+
+	log "github.com/golang/glog"
+)
+
+func assert(t *testing.T, fact string, condition bool) {
```
ConnorDoyle: @jdef this `assert` definition is not idiomatic -- any suggestion for me here?  Leaning toward removing this and updating the call sites to use the expanded form.

ConnorDoyle: Removed.

ConnorDoyle:
@jdef @joerg84 PTAL

ConnorDoyle:
Thanks for the update @joerg84.

pkg/cloud/mesos/mesos_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	reader := bytes.NewBufferString(configString)
+
+	mesosCloud, err := newMesosCloud(reader)
+
+	if err != nil {
+		t.Fatalf("Creating a new Mesos cloud provider with a custom config does not yield an error: %#v", err)
+	}
+
+	if mesosCloud.client.httpClient.Timeout != time.Duration(500)*time.Millisecond {
```
ConnorDoyle: should reference some const variables here instead of repeating what's in the global default

pkg/cloud/mesos/config.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+const (
+	DefaultMesosMaster       = "localhost:5050"
+	DefaultHttpClientTimeout = time.Duration(10) * time.Second
+	DefaultStateCacheTTL     = time.Duration(5) * time.Second
+)
+
+var (
+	config *Config
```
ConnorDoyle: make this a member variable, not global

pkg/cloud/mesos/config_test.go
```diff
@@ -0,0 +1,59 @@
...
+func Test_createDefaultConfig(t *testing.T) {
+	defer log.Flush()
+
+	config := createDefaultConfig()
+
+	if config.MesosMaster != "localhost:5050" {
+		t.Fatalf("Default config has the expected MesosMaster value")
+	}
+
+	if config.MesosHttpClientTimeout.Duration != time.Duration(10)*time.Second {
```
ConnorDoyle: DRY: reference const variables here

pkg/cloud/mesos/config_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if err != nil {
+		t.Fatalf("Reading configuration does not yield an error: %#v", err)
+	}
+
+	if getConfig().MesosMaster != "leader.mesos:5050" {
+		t.Fatalf("Parsed config has the expected MesosMaster value")
+	}
+
+	if getConfig().MesosHttpClientTimeout.Duration != time.Duration(500)*time.Millisecond {
```
ConnorDoyle: DRY: use const values

pkg/cloud/mesos/client_test.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+}
+
+// test mesos.listSlaves
+func Test_listSlaves(t *testing.T) {
+	defer log.Flush()
+	md := FakeMasterDetector{}
+	httpServer, httpClient, httpTransport := makeHttpMocks()
+	defer httpServer.Close()
+	cacheTTL, _ := time.ParseDuration("500ms")
```
ConnorDoyle: cacheTTL = 500 * time.Millisecond

pkg/cloud/mesos/client_test.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+}
+
+// test mesos.clusterName
+func Test_clusterName(t *testing.T) {
+	defer log.Flush()
+	md := FakeMasterDetector{}
+	httpServer, httpClient, httpTransport := makeHttpMocks()
+	defer httpServer.Close()
+	cacheTTL, _ := time.ParseDuration("500ms")
```
ConnorDoyle: 500*time.Millisecond

pkg/cloud/mesos/config.go
```diff
@@ -0,0 +1,79 @@
...
+	Mesos_Cloud Config
+}
+
+type Config struct {
+	MesosMaster            string          `gcfg:"mesos-master"`
+	MesosHttpClientTimeout WrappedDuration `gcfg:"http-client-timeout"`
+	StateCacheTTL          WrappedDuration `gcfg:"state-cache-ttl"`
+}
+
+type WrappedDuration struct {
```
jdef: it looks like @sttts wants the same thing as part of #246. should probably put this is a common place. also, should probably use an anonymous member here instead. for an example, see how k8s wraps Time: https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/util/time.go

ConnorDoyle: Two questions:

1. This will eventually (hopefully soon) be moved upstream.  I could put it in a common location at that time.
1. Wrapping every function we want to call on the underlying thing as in the linked example seems silly to me.  Is is really necessary?  If calling code want to invoke functions that we didn't think to wrap, they will have to go to the underlying member anyhow, or add more wrappers.  Then again, maybe I didn't understand what part of the linked code you were referring to.

jdef: Sorry I should have explained myself. We don't want to duplicate all the methods like they do. I was trying to point out that:
```go
type Duration struct {
   time.Duration
}
```
is more idiomatic than:
```go
type WrappedDuration struct {
    Duration time.Duration
}
```

If we want to eliminate the `.Duration` but keep access to funcs on `time.Duration` then we can declare this way:
```go
type Duration struct {
   *time.Duration
}
```

So, now we can say something like this:
```go
func doSomething(d Duration) {
   fmt.Printf("number of seconds in d is %d", d.Seconds())
}
```

pkg/scheduler/service/service.go
```diff
@@ -547,7 +547,7 @@ func (s *SchedulerServer) bootstrap(hks hyperkube.Interface) (*ha.SchedulerProce
 		ReconcileCooldown: s.ReconcileCooldown,
 	})
 
-	masterUri := kmcloud.MasterURI()
+	masterUri := kmcloud.CloudProvider.MasterURI()
```
jdef: now that we're using the cloud provider to get this URI, we probably need to make sure it's initialized before we attempt to access it.

https://github.com/GoogleCloudPlatform/kubernetes/blob/release-0.14/pkg/cloudprovider/plugins.go#L66

jdef:
Like the changes. I think there's an initialization bug (commented). Has this seen any integration testing?

ConnorDoyle:
No integration testing with the latest iteration.  We spun up some pods with the changes as of cbc67a2.

pkg/scheduler/service/service.go
ConnorDoyle: Good call, thought about this when implementing and then subsequently forgot about it.  I'll change the factory function we supply when registering to be idempotent to avoid initializing multiple times.

ConnorDoyle:
I'm also adding a few final tests to boost coverage in `mesos.go`.

jdef:
SGTM

ConnorDoyle:
Ping @jdef -- added some synchronization mechanisms around cloud provider initialization to address your previous comment.  Let me know if there's anything else that sticks out, including whether you'd prefer to see the commits squashed.

ConnorDoyle:
Test coverage for the cloud provider is at 76.3%.

pkg/scheduler/service/service.go
```diff
@@ -547,7 +547,8 @@ func (s *SchedulerServer) bootstrap(hks hyperkube.Interface) (*ha.SchedulerProce
 		ReconcileCooldown: s.ReconcileCooldown,
 	})
 
-	masterUri := kmcloud.MasterURI()
+	kmcloud.AwaitInitialization() // Blocking
```
jdef: do we need to invoke cloudprovider.InitCloudProvider(...) before waiting? Not sure that this blocking call will ever return unless we do.

README.md
```diff
@@ -85,34 +85,49 @@ If you are running an older version of etcd you may want to consider an upgrade.
...
   --port=8888 \
   --cloud_provider=mesos
+  --cloud_config=./mesos-cloud.conf
 
 $ ./bin/km controller-manager \
   --master=$servicehost:8888 \
-  --mesos_master=${mesos_master}
+ --cloud_config=./mesos-cloud.conf
 
 $ ./bin/km scheduler \
```
jdef: doesn't the scheduler need a `--cloud_config` now? how else will it find the mesos master?

README.md
```diff
@@ -85,34 +85,49 @@ If you are running an older version of etcd you may want to consider an upgrade.
...
   --etcd_servers=http://${servicehost}:4001 \
   --portal_net=10.10.10.0/24 \
   --port=8888 \
   --cloud_provider=mesos
+  --cloud_config=./mesos-cloud.conf
 
 $ ./bin/km controller-manager \
   --master=$servicehost:8888 \
-  --mesos_master=${mesos_master}
+ --cloud_config=./mesos-cloud.conf
```
jdef: missing a space

pkg/cloud/mesos/mesos.go
```diff
@@ -14,67 +15,118 @@ import (
...
 )
 
 func init() {
 	cloudprovider.RegisterCloudProvider(
 		PluginName,
-		func(conf io.Reader) (cloudprovider.Interface, error) {
-			return newMesosCloud()
+		func(configReader io.Reader) (cloudprovider.Interface, error) {
+			provider, err := newMesosCloud(configReader)
+			if CloudProvider != nil {
```
jdef: should this check come before the line above it?

pkg/cloud/mesos/mesos.go
```diff
@@ -14,67 +15,118 @@ import (
...
 }
 
 type MesosCloud struct {
 	client *mesosClient
+	config *Config
+}
+
+// Blocks until the Mesos cloud provider has been initialized via the factory
+// function defined at registration time.
+func AwaitInitialization() {
```
jdef: maybe this would be more aptly named CheckInitialized():
```go
func CheckInitialized() {
  select {
    case <-initialized:
      return
    default:
      //programmer error
      panic("missing prior call to cloudprovider.InitCloudProvider for 'mesos' cloud provider")
  }
}

File: pkg/cloud/mesos/mesos.go (hunk 17 above)
Comment (ConnorDoyle) [inline pkg/cloud/mesos/mesos.go]: yep!

File: README.md (hunk 16 above)
Comment (ConnorDoyle) [inline README.md]: will fix

File: README.md (hunk 15 above)
Comment (ConnorDoyle) [inline README.md]: @jdef I'm more inclined at this point to reintroduce a --mesos_master flag to the Scheduler in order to avoid this awkward and unnecessary coupling to the cloud provider.  Will update the PR to do that shortly -- should also be able to completely remove the AwaitInitialization business (unless you think it might be useful elsewhere).

Comment (jdef) [inline README.md]: SGTM. The scheduler shouldn't need any other part of the cloud provider,
and the underlying mesos-go lib handles master detection.

**Comment by ConnorDoyle** [comment]
@jdef -- pushed that final cleanup for the scheduler and rebased.

File: pkg/scheduler/service/service.go [hunk 19]
```diff
@@ -174,8 +175,9 @@ func (s *SchedulerServer) addCoreFlags(fs *pflag.FlagSet) {
...
 	fs.BoolVar(&s.AllowPrivileged, "allow_privileged", s.AllowPrivileged, "If true, allow privileged containers.")
 	fs.StringVar(&s.ClusterDomain, "cluster_domain", s.ClusterDomain, "Domain for this cluster.  If set, kubelet will configure all containers to search this domain in addition to the host's search domains")
-	fs.Var(&s.ClusterDNS, "cluster_dns", "IP address for a cluster DNS server.  If set, kubelet will configure all containers to use this for DNS resolution in addition to the host's DNS servers")
+	fs.Var(&s.ClusterDNS, "cluster_dns", "IP address for a cluster DNS server. If set, kubelet will configure all containers to use this for DNS resolution in addition to the host's DNS servers")
 
+	fs.StringVar(&s.MesosMaster, "mesos_master", s.MesosUser, "Location of the Mesos master. The format is a comma-delimited list of of hosts like zk://host1:port,host2:port/mesos. If using ZooKeeper, pay particular attention to the leading zk:// and trailing /mesos! If not using ZooKeeper, standard URLs like http://localhost are also acceptable.")
```
jdef: The default value here should be `s.MesosMaster` (and then initialize this in the `NewSchedulerServer()` func) or `defaultMesosMaster`; not `s.MesosUser`;

README.md
```diff
@@ -85,34 +85,49 @@ If you are running an older version of etcd you may want to consider an upgrade.
...
+  --cloud_config=./mesos-cloud.conf
 
 $ ./bin/km controller-manager \
   --master=$servicehost:8888 \
-  --mesos_master=${mesos_master}
+  --cloud_config=./mesos-cloud.conf
 
 $ ./bin/km scheduler \
   --address=${servicehost} \
-  --mesos_master=${mesos_master} \
```
jdef: need to add back this flag

pkg/scheduler/service/service.go
ConnorDoyle: Cool, done.

README.md
ConnorDoyle: done.

pkg/cloud/mesos/mesos.go
```diff
@@ -14,67 +15,111 @@ import (
 )
 
 var (
-	PluginName = "mesos"
+	PluginName    = "mesos"
+	CloudProvider *MesosCloud
 
+	initialized         = make(chan struct{})
```
jdef: can we remove this now?

ConnorDoyle: Yep!

jdef:
:shipit: once comments are addressed

ConnorDoyle:
Thanks for the reviews!  Will ship when CI is done.

ConnorDoyle:
:confetti_ball: :tada: :balloon: :rocket:

Issue #194, opened 2015-03-30
refactor slave tracking in scheduler


jdef:
the `KubernetesScheduler` maintains a slave index and uses it's "master lock" to control access to that index. The index really should be broken out into its own registry, similar to what's been done for offers and pod tasks. Then we can probably eliminate the big ugly lock at the top-level scheduler (or at least get closer to that point).

jdef:
I already ripped out the slave's embedded offer list as part of #207. Should be trivial at this point to write a slaveStorage struct and move the lock in there.

jdef:
this issue needs review .. anything left?

karlkfi:
I think this is done, since https://github.com/mesosphere/kubernetes-mesos/pull/220 has long since been merged. Feel free to make a new ticket if you don't agree, @jdef

jdef:
sgtm

PR #220, opened 2015-04-13
Refactor slave tracking in scheduler

joerg84:
#194

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -56,6 +56,44 @@ func newSlave(hostName string) *Slave {
 	}
 }
 
+type slaveStorage struct {
+	sync.Mutex
+	slaves map[string]*Slave // SlaveID => slave.
+}
+
+func newSlaveStorage() *slaveStorage {
+	return &slaveStorage{
+		slaves: make(map[string]*Slave),
+	}
+}
+
+// Create a mapping between a slaveID and slave if not existing.
+func (self *slaveStorage) checkAndAdd(slaveId, slaveHostname string) {
+	self.Lock()
+	defer self.Unlock()
+	_, exists := self.slaves[slaveId]
+	if !exists {
+		self.slaves[slaveId] = newSlave(slaveHostname)
+	}
+}
+
+func (self *slaveStorage) getSlaveIds() []string {
+	self.Lock()
+	defer self.Unlock()
+	slaveIds := make([]string, 0, len(self.slaves))
+	for slaveID, _ := range self.slaves {
+		slaveIds = append(slaveIds, slaveID)
+	}
+	return slaveIds
+}
+
+func (self *slaveStorage) getSlave(slaveId string) (*Slave, bool) {
+	self.Lock()
+	defer self.Unlock()
+	slave, exists := self.slaves[slaveId]
+	return slave, exists
+}
+
 type PluginInterface interface {
 	// the apiserver may have a different state for the pod than we do
 	// so reconcile our records, but only for this one pod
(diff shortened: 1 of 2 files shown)


ConnorDoyle:
:+1: Looking cleaner!

jdef:
This cleanup is long overdue. Thanks for implementing it. Couple of suggestions:

1. please run `make format` or else individually `go fmt` the source code before you commit
1. `exists()` and `getSlave()` are redundant. recommend eliminating `exists()`, clients can query the same via `_, found := slaveStorage.getSlave(someId)` instead. it's more idiomatic.

pkg/scheduler/scheduler.go
```diff
@@ -230,10 +276,7 @@ func (k *KubernetesScheduler) InstallDebugHandlers(mux *http.ServeMux) {
 		func() {
```
jdef: can actually eliminate this anonymous func() since it's only there to provide context for lock/unlock. instead of using a chan (get rid of that too), just set `slaves := k.slaves.getSlaveIds()` directly.

pkg/scheduler/scheduler.go
```diff
@@ -388,7 +428,7 @@ func (k *KubernetesScheduler) StatusUpdate(driver bindings.SchedulerDriver, task
 		}
 		k.RLock()
```
jdef: can safely remove this RLock/RUnlock since the only thing they were guarding is now protected via `slaveStorage`.

pkg/scheduler/scheduler.go
```diff
@@ -335,10 +378,7 @@ func (k *KubernetesScheduler) ResourceOffers(driver bindings.SchedulerDriver, of
 	k.offers.Add(offers)
 	for _, offer := range offers {
 		slaveId := offer.GetSlaveId().GetValue()
-		_, exists := k.slaves[slaveId]
-		if !exists {
-			k.slaves[slaveId] = newSlave(offer.GetHostname())
-		}
+		k.slaves.checkAndAdd(slaveId, offer.GetHostname())
```
jdef: you can safely remove the lock/unlock that appears a few lines back -- it's no longer needed

pkg/scheduler/scheduler.go
```diff
@@ -845,11 +885,7 @@ func (ks *KubernetesScheduler) recoverTasks() error {
 		defer ks.Unlock()
```
jdef: you can safely remove the calls to lock/unlock here

joerg84: I actually planned to remove the redundant locks in a second round. Is that ok or do you prefer this in the same patch?

jdef: Second round is fine. I thought I saw that you had already started, so I
was just pointing out the rest.

On Mon, Apr 13, 2015 at 7:11 AM, Joerg Schad <notifications@github.com>
wrote:

> In pkg/scheduler/scheduler.go
> <https://github.com/mesosphere/kubernetes-mesos/pull/220#discussion_r28229983>
> :
>
> > @@ -845,11 +885,7 @@ func (ks *KubernetesScheduler) recoverTasks() error {
> >  		defer ks.Unlock()
>
> I actually planned to remove the redundant locks in a second round. Is
> that ok or do you prefer this in the same patch?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mesosphere/kubernetes-mesos/pull/220/files#r28229983>.
>



-- 
James DeFelice
585.241.9488 (voice)
650.649.6071 (fax)

jdef:
LGTM

Issue #191, opened 2015-03-28
prototype using a make.go file instead of Makefile for builds


jdef:
examples:
* https://camlistore.googlesource.com/camlistore/+/master/make.go
* https://gist.github.com/rbucker/2f362aeb0ddec24e8b8f

jdef:
we're using the upstream build process now

PR #227, opened 2015-04-17
Makefile written in GO

ovisan:
this is the initial commit that has ```install``` and ```all``` commands working
https://github.com/mesosphere/kubernetes-mesos/issues/191

### DIFF
--- a/makefile.go
+++ b/makefile.go
@@ -1,0 +1,208 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path"
+	"path/filepath"
+)
+
+var (
+	kube_go_package string
+	k8sm_go_package string
+	k8_cmd          []string
+	tempDirs        []string
+	cmd_dirs        []string
+	lib_dirs        []string
+	target          = "./target"
+)
+
+func dieIf(err error) {
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func run(dir, command string, args ...string) error {
+	cmd := exec.Command(command, args...)
+	cmd.Dir = dir
+
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	err := cmd.Run()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// a walk path function to search for all *.go files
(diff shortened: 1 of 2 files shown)


makefile.go
```diff
@@ -0,0 +1,220 @@
...
+	target          = "./target"
+)
+
+func run(dir, command string, args ...string) (output string, err error) {
+	cmd := exec.Command(command, args...)
+	cmd.Dir = dir
+
+	var out bytes.Buffer
+	var stderr bytes.Buffer
+	cmd.Stdout = &out
```
jdef: why capture the output/error streams instead of just configuring cmd to use `os.Stdout` and `os.Stdin`?

makefile.go
```diff
@@ -0,0 +1,220 @@
...
+	KUBE_GO_PACKAGE string
+	K8SM_GO_PACKAGE string
+	K8_CMD          []string
+	tempDirs        []string
+	CMD_DIRS        []string
+	LIB_DIRS        []string
+	target          = "./target"
+)
+
+func run(dir, command string, args ...string) (output string, err error) {
```
jdef: the output string is never actually used anywhere, should probably kill it and just return error

makefile.go
```diff
@@ -0,0 +1,220 @@
...
+	"os/exec"
+	"path"
+	"path/filepath"
+)
+
+var (
+	KUBE_GO_PACKAGE string
+	K8SM_GO_PACKAGE string
+	K8_CMD          []string
+	tempDirs        []string
```
jdef: smells funny to have this as a global

jdef:
Thanks for the PR. This is a nice first step towards breaking away from the Makefile. I'd recommend adding something like `func dieIf(err error) { if err != nil { log.Fatal(err) } }` that you can wrap around calls that simply return `error` -- saves on some of the boilerplate.

ConnorDoyle:
Could you also make the associated changes to the build instructions in the README and elsewhere?

makefile.go
ovisan: I did it like this because later on it's maybe needed to parse the output for further actions, but it can be changed back later.

ovisan:
You are welcome :-) 
The README contains now the build instructions and the code has been modified with your suggestions.

PR #232, opened 2015-04-28
removed reference to master_mesos env variable

joerg84:
Follow Up to #210 
Quick test looks good, but I will replay the entire tutorial with these changes tomorrow.
Note: we should also update the GCE tutorial once these changes hit upstream/once we tag another release.


joerg84:
Pushed updated version directly to upstream-cloud-provider-prep brach.

PR #231, opened 2015-04-26
Add test coverage for the offers package

sttts:

PR #243, opened 2015-05-02
Add test coverage for the offers package

sttts:
Replaces #231.

### DIFF
--- a/pkg/offers/offers_test.go
+++ b/pkg/offers/offers_test.go
@@ -10,6 +10,32 @@ import (
 	"github.com/mesosphere/kubernetes-mesos/pkg/proc"
 )
 
+func TestExpiredOffer(t *testing.T) {
+	t.Parallel()
+
+	ttl := 2 * time.Second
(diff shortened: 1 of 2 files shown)


pkg/offers/offers_test.go
```diff
@@ -47,6 +73,212 @@ func TestTimedOffer(t *testing.T) {
...
+	})
+
+	// Add hostname1 offer
+	id := util.NewOfferID("foo")
+	o := &mesos.Offer{Id: id, Hostname:&hostname1}
+	storage.Add([]*mesos.Offer{o})
+
+	// listener1 is notified by closing channel
+	select {
+	case _, more := <-listener1:
```
jdef: (very) small nit: "more" is an interesting choice. i think that if <-someChan returns `_, true` then it doesn't necessarily mean that there's more data coming down the chan, but that it just hasn't closed yet. people tend to use `ok` or `found` for instead.

jdef:
LGTM - thanks for the PR!

PR #245, opened 2015-05-03
Added scheduler tests, Updated Driver Mock

joerg84:

PR #246, opened 2015-05-03
Add --scheduler-config parameter for low-level scheduler settings

### LINKED ISSUE #237: Make scheduler behavior more configurable
There are a number of constant "default" values that cannot currently be configured in the scheduler (see the values in [scheduler.go](https://github.com/mesosphere/kubernetes-mesos/blob/master/pkg/scheduler/scheduler.go#L36)).

We could add flags, read them from a config file (or both).  For reference, here's how the cloud provider will read config from a file with gcfg: https://github.com/mesosphere/kubernetes-mesos/blob/437d0bc4c8c48f3220d589e9ea82fde3edfa4485/pkg/cloud/mesos/config.go.

CC @sttts

sttts:
Fixes #237

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,45 @@ $ export KUBERNETES_MASTER=http://${servicehost}:8888
 You can increase the verbosity of the logging for the API server, scheduler, and/or the controller-manager by including, for example, `--v=2`.
 This can be very helpful while debugging.
 
+#### Scheduler Configuration
+
+The scheduler configuration can be fine-tuned using an ini-style configuration file. The filename is passed via `--scheduler_config` to the `km scheduler` command. The following settings are the default:
+
+```
+[scheduler]
+; intended to be mesos FrameworkInfo.Name
+info-name = Kubernetes
+
+; duration an offer is viable, prior to being expired
+offer-ttl = 5s
+
+; duration an expired offer lingers in history
+offer-linger-ttl = 2m
+
+; duration between offer listener notifications
+listener-delay = 1s
+
+; size of the pod updates channel
+updates-backlog = 2048
+
+; interval we update the frameworkId stored in etcd
+framework-id-refresh-interval = 30s
+
+; wait this amount of time after initial registration before attempting
+; implicit reconciliation
+initial-implicit-reconciliation-delay = 15s
+
+; interval in between internal task status checks/updates
+explicit-reconciliation-max-backoff = 2m
+
+; waiting period after attempting to cancel an ongoing reconciliation
+explicit-reconciliation-abort-timeout = 30s
+
+initial-pod-backoff = 1s
+max-pod-backoff = 60s
+http-handler-timeout = 10s
+```
+
 ###Launch a Pod
 
 Assuming your framework is running on `${KUBERNETES_MASTER}`, then:
--- a/README.md
+++ b/README.md
@@ -126,42 +126,6 @@ $ export KUBERNETES_MASTER=http://${servicehost}:8888
 You can increase the verbosity of the logging for the API server, scheduler, and/or the controller-manager by including, for example, `--v=2`.
 This can be very helpful while debugging.
 
-#### Scheduler Configuration
-
-The scheduler configuration can be fine-tuned using an ini-style configuration file. The filename is passed via `--scheduler_config` to the `km scheduler` command. The following settings are the default:
-
-```
-[scheduler]
-; duration an offer is viable, prior to being expired
-offer-ttl = 5s
-
-; duration an expired offer lingers in history
-offer-linger-ttl = 2m
-
-; duration between offer listener notifications
-listener-delay = 1s
-
-; size of the pod updates channel
-updates-backlog = 2048
-
-; interval we update the frameworkId stored in etcd
-framework-id-refresh-interval = 30s
-
-; wait this amount of time after initial registration before attempting
-; implicit reconciliation
-initial-implicit-reconciliation-delay = 15s
-
-; interval in between internal task status checks/updates
-explicit-reconciliation-max-backoff = 2m
-
-; waiting period after attempting to cancel an ongoing reconciliation
-explicit-reconciliation-abort-timeout = 30s
-
-initial-pod-backoff = 1s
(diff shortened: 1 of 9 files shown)


sttts:
Waiting for #245 to add tests on top of that regarding the new config file.

pkg/scheduler/config/config.go
```diff
@@ -1,6 +1,89 @@
...
+	DefaultInitialPodBackoff                  = 1 * time.Second
+	DefaultMaxPodBackoff                      = 60 * time.Second
+	DefaultHttpHandlerTimeout                 = 10 * time.Second
 )
+
+// Example scheduler configuration file:
+//
+// [scheduler]
+//  info-name        = Kubernetes
+//	offer-ttl        = 5s
```
jdef: intentional whitespace offset?

sttts: Nope. Shouldn't be there.

pkg/scheduler/config/config.go
```diff
@@ -1,6 +1,89 @@
...
+	FrameworkIdRefreshInterval         WrappedDuration `gcfg:"listener-delay"`
+	InitialImplicitReconciliationDelay WrappedDuration `gcfg:"initial-implicit-reconciliation-delay"`
+	ExplicitReconciliationMaxBackoff   WrappedDuration `gcfg:"explicit-reconciliantion-max-backoff"`
+	ExplicitReconciliationAbortTimeout WrappedDuration `gcfg:"explicit-reconciliantion-abort-timeout"`
+	InitialPodBackoff                  WrappedDuration `gcfg:"initial-pod-backoff"`
+	MaxPodBackoff                      WrappedDuration `gcfg:"max-pod-backoff"`
+	HttpHandlerTimeout                 WrappedDuration `gcfg:"http-handler-timeout"`
+}
+
+type WrappedDuration struct {
```
jdef: I think it's more idiomatic to use an anonymous member here. For an example see how the k8s team wrapped Time: https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/util/time.go

@ConnorDoyle also wants to wrap Duration as part of #210 -- we should probably have a common wrapper that gets reused.

pkg/scheduler/scheduler.go
```diff
@@ -744,9 +735,11 @@ type Reconciler struct {
 	implicit chan struct{}   // send an empty struct to trigger implicit reconciliation
 	done     <-chan struct{} // close this when you want the reconciler to exit
 	cooldown time.Duration
+	explicitReconciliationAbortTimeout time.Duration
 }
 
-func newReconciler(doer proc.Doer, action ReconcilerAction, cooldown time.Duration, done <-chan struct{}) *Reconciler {
+func newReconciler(doer proc.Doer, action ReconcilerAction, cooldown time.Duration,
```
jdef: looks like `Reconciler.explicitReconciliationAbortTimeout` is never assigned

jdef: also, more idiomatic to declare, for example, `f(a,b int)` instead of `f(a int, b int)` .. w/ respect to the time.Duration params

pkg/scheduler/service/service.go
```diff
@@ -68,6 +69,7 @@ const (
 	defaultReconcileInterval = 300    // 5m default task reconciliation interval
 	defaultReconcileCooldown = 15 * time.Second
 	defaultHttpBindInterval  = 5 * time.Second
```
jdef: might want make the default http bind interval part of the scheduler fine tuning config file.

README.md
```diff
@@ -126,6 +126,42 @@ $ export KUBERNETES_MASTER=http://${servicehost}:8888
 You can increase the verbosity of the logging for the API server, scheduler, and/or the controller-manager by including, for example, `--v=2`.
 This can be very helpful while debugging.
 
+#### Scheduler Configuration
```
jdef: this is great information to document, but I'd rather see it in some kind of "tuning" document vs. in the main readme.

TUNING.md
```diff
@@ -0,0 +1,41 @@
+# Tuning
+
+There are more settings which can be customized to adapt k8sm's behaviour to you environment. Be warned though that some them are pretty low-level and one has to know the inner workings of k8sm to find sensible values. Moreover, these
```
jdef: nice

TUNING.md
```diff
@@ -0,0 +1,41 @@
...
+setting may change or even disappear from version to version without further notice.
```
jdef: s/setting/settings/

pkg/scheduler/scheduler.go
```diff
@@ -331,13 +321,14 @@ func (k *KubernetesScheduler) onInitialRegistration(driver bindings.SchedulerDri
...
+	k.reconciler = newReconciler(k.asRegisteredMaster, k.makeCompositeReconciler(r1, r2),
+		k.reconcileCooldown, k.schedcfg.ExplicitReconciliationAbortTimeout.Duration, k.terminate)
 	go k.reconciler.Run(driver)
 
 	if k.reconcileInterval > 0 {
 		ri := time.Duration(k.reconcileInterval) * time.Second
-		time.AfterFunc(initialImplicitReconciliationDelay, func() { runtime.Until(k.reconciler.RequestImplicit, ri, k.terminate) })
-		log.Infof("will perform implicit task reconciliation at interval: %v after %v", ri, initialImplicitReconciliationDelay)
+		time.AfterFunc(k.schedcfg.InitialImplicitReconciliationDelay.Duration, func() { runtime.Until(k.reconciler.RequestImplicit, ri, k.terminate) })
+		log.Infof("will perform implicit task reconciliation at interval: %v after %v", ri, k.schedcfg.InitialImplicitReconciliationDelay)
```
jdef: WrappedDuration doesn't have a String() conversion func, so you should probably add a `.Duration` so that the logging is better

jdef:
LGTM overall, couple of minor nits. also, please run `make format`. ready for merge after these things are addressed.

sttts:
addressed all remaining comments.

PR #248, opened 2015-05-04
Update development docs and Makefile for Mac OS X building

karlkfi:
Building works on Mac OS 10.10 with this minor change. 
I also documented the prerequisites, installed using homebrew.


jdef:
thanks for the PR

Issue #252, opened 2015-05-05
Mesos Authentification: Auth file not always required


joerg84:
A Credentials/Secret file is not always required, as for example with Kerberos Authentication. 
Will discuss further with @tillt.

if s.MesosAuthPrincipal != "" {
		info.Principal = proto.String(s.MesosAuthPrincipal)
		if s.MesosAuthSecretFile == "" {
			return nil, nil, errors.New("authentication principal specified without the required credentials file")
		}
		secret, err := ioutil.ReadFile(s.MesosAuthSecretFile)
		if err != nil {
			return nil, nil, err
		}
		cred = &mesos.Credential{
			Principal: proto.String(s.MesosAuthPrincipal),
			Secret:    secret,
		}
	}

PR #253, opened 2015-05-05
Added buildFramework test.

joerg84:


joerg84:
I will add more details for the credential testing with #252

PR #254, opened 2015-05-05
Wildcard ports

### LINKED ISSUE #59: services should not require pods and pod-templates to declare explicit hostPort's

jdef:
- resolves #59 
- fixes bugs in proxy deployment script

TODO:
- [x] a few more unit tests, specially around the new code in endpoint_controller

### DIFF
--- a/examples/service.json
+++ b/examples/service.json
@@ -3,7 +3,7 @@
   "kind": "Service",
   "apiVersion": "v1beta1",
   "port": 8000,
-  "containerPort": 80,
+  "containerPort": "http",
   "selector": {
     "name": "foo"
   }
--- a/examples/service.json
+++ b/examples/service.json
@@ -3,6 +3,7 @@
   "kind": "Service",
   "apiVersion": "v1beta1",
   "port": 8000,
+  "containerPort": 80,
   "selector": {
     "name": "foo"
   }
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -7,9 +7,10 @@ Kubernetes pod container specifications identify two types of ports, container p
 container ports are allocated from the network namespace of the pod, which is independent from that of the host, whereas;
 host ports are allocated from the network namespace of the host.
 The k8sm scheduler recognizes the declared host ports of each container in a pod/task and for each such port, attempts to allocate it from the offered ports listed in mesos resource offers.
+If no host port is declared, then the scheduler may choose any port from the offered ports ranges.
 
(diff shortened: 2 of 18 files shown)


jdef:
@ConnorDoyle @joerg84 PTAL

joerg84:
Ship It.

examples/service.json
```diff
@@ -3,6 +3,7 @@
   "kind": "Service",
   "apiVersion": "v1beta1",
   "port": 8000,
+  "containerPort": "http",
```
ConnorDoyle: Is this value supposed to be an integer?

jdef: k8s has a nifty IntOrString type. that's what the type of containerPort is:

https://github.com/GoogleCloudPlatform/kubernetes/blob/release-0.14/pkg/api/v1beta1/types.go#L739

ConnorDoyle: Cool, thanks.

ConnorDoyle:
+1, LGTM.  Nice tests!

PR #258, opened 2015-05-06
mesos-go, zookeeper rebase

### LINKED ISSUE #59: services should not require pods and pod-templates to declare explicit hostPort's

### LINKED ISSUE #255: rebase to latest mesos-go
need to pick up zk detection bug fixes

jdef:
- rebase to latest mesos-go
- rebase to latest samuel/go-zookeeper

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -690,35 +690,35 @@
 		},
 		{
 			"ImportPath": "github.com/mesos/mesos-go/auth",
-			"Rev": "6a19badfefc9ddec8c0c8c6ef39aec9fbed78f90"
+			"Rev": "4b1767c0dfc51020e01f35da5b38472f40ce572a"
 		},
 		{
 			"ImportPath": "github.com/mesos/mesos-go/detector",
-			"Rev": "6a19badfefc9ddec8c0c8c6ef39aec9fbed78f90"
+			"Rev": "4b1767c0dfc51020e01f35da5b38472f40ce572a"
 		},
 		{
 			"ImportPath": "github.com/mesos/mesos-go/executor",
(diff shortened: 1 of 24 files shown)


jdef:
rebuilt the dcos image from this branch, seems to work well enough.

PR #260, opened 2015-05-06
initial support for framework weburl

### LINKED ISSUE #259: framework web_uri should be set when deploying on dcos
in the dcos "service health" pane Marathon gets a link to its WebUI but Kubernetes does not. The scheduler should probably fill in the web_uri field of FrameworkInfo, but it'll need to be configured via an optional flag since the apiserver actually hosts the UI that we'd be linking to.

jdef:
- resolves #259


jdef:
I'd add a unit test but I think @joerg84 is already testing the buildFrameworkInfo func in another branch

jdef:
ready for merge, PTAL

ConnorDoyle:
:shipit:

PR #261, opened 2015-05-07
Cleanup of scheduler uid

karlkfi:
- Moved uuid.New() use to uid.Generate()
- Avoid pointers, treat as an immutable object after construction
- Move some scheduler logic out of the constructor


pkg/scheduler/service/service.go
```diff
@@ -473,14 +473,25 @@ func (s *SchedulerServer) awaitFailover(schedulerProcess schedulerProcessInterfa
 
 func validateLeadershipTransition(desired, current string) {
 	log.Infof("validating leadership transition")
-	d := uid.Parse(desired).Group()
-	c := uid.Parse(current).Group()
-	if d == 0 {
+	desiredUID, err := uid.Parse(desired)
+	if err != nil {
+		log.Fatalf("Failed to parse desired scheduler UID %q: %v", desired, err)
```
jdef: Would you mind adding a func-level comment (one that I should have added but did not)? something along the lines of "this func may terminate the program if leadership transition validation fails. this is purposeful and attempts to thwart split-brain scenarios. it *should* be safe to terminate this way because there should be some sort of external service monitoring process in place that will restart the scheduler after it terminates".

pkg/scheduler/uid/uid.go
```diff
@@ -15,55 +15,46 @@ type UID struct {
...
-		log.Errorf("illegal UID group %q: %v", parts[0], err)
-		return nil
+		return UID{}, fmt.Errorf("invalid UID group %q: %v", parts[0], err)
 	}
-	if parts[1] == "" {
-		log.Errorf("missing UID name: %q", ser)
-		return nil
+	name := parts[1]
+	if name == "" {
+		return UID{}, fmt.Errorf("invalid UID name %q (expected non-empty)", name)
```
jdef: why log name here if it's empty?

pkg/scheduler/uid/uid.go
```diff
@@ -15,55 +15,46 @@ type UID struct {
 	ser   string
 }
 
-func New(group uint64, name string) *UID {
-	if name == "" {
-		name = uuid.New()
-	}
-	return &UID{
+func New(group uint64, name string) UID {
+	return UID{
```
jdef: why eliminate the check for an empty name? maybe return an error in that case? it was my intent that name would never be empty

pkg/scheduler/uid/uid.go
```diff
@@ -15,55 +15,46 @@ type UID struct {
...
+func (u UID) String() string {
+	return u.ser
 }
 
-func Parse(ser string) *UID {
+func Parse(ser string) (UID, error) {
 	parts := strings.SplitN(ser, "_", 2)
 	if len(parts) != 2 {
-		return nil
+		return UID{}, errors.New("invalid UID format (expected <uint64>_<string>)")
```
jdef: maybe "64bit-hex" instead of "uint64"?

pkg/scheduler/scheduler.go
```diff
@@ -160,39 +160,26 @@ type Config struct {
 
 // New create a new KubernetesScheduler
 func New(config Config) *KubernetesScheduler {
+	executorId := config.Executor.ExecutorId.GetValue()
+	executorUID, err := uid.Parse(executorId)
+	if err != nil {
```
jdef: please return the error instead of fatally exiting here. i'm really trying to limit the places where fatally exiting the program is OK. i'm generally not convinced that doing so in a constructor is OK.

pkg/scheduler/service/service.go
```diff
@@ -473,14 +473,25 @@ func (s *SchedulerServer) awaitFailover(schedulerProcess schedulerProcessInterfa
...
+
+	desiredGroup := desiredUID.Group()
+	if desiredGroup == 0 {
 		// should *never* happen, but..
 		log.Fatalf("illegal scheduler UID: %q", desired)
 	}
-	if d != c && c != 0 {
-		log.Fatalf("desired scheduler group (%x) != current scheduler group (%x)", d, c)
+
+	currentUID, err := uid.Parse(desired)
```
jdef: you've changed the semantics here. the 0 current-group (c) was allowed before. Parse returns an error instead of nil now and the code fatally exits on that condition. so this func will behave differently. desired action via truth table:
```
d | c | result
---------------
0 | 0 | fail
0 | 1 | fail
0 | 2 | fail
1 | 0 | c was nil (no current leader); OK
1 | 1 | d == c; OK
1 | 2 | d != c && c != 0; fail
2 | 0 | c was nil (no current leader); OK
2 | 1 | d != c && c != 0; fail
2 | 2 | d == c; OK
```

jdef: may want a unit test for this

pkg/scheduler/uid/uid.go
karlkfi: It's logging the whole id, not just the name. Tho I see how the colon following "name" might be confusing.

pkg/scheduler/uid/uid.go
karlkfi: I moved generation to its own Generate method. I like to avoid optional parameters that need zero values passed because it makes the code harder to read. I could replace the default behavior with a fatal panic, however, since that's the idiomatic way to fail in a constructor due to an invalid usage error. If we need the constructor not to explode, the normal solution is to either validate before using it or have another static constructor that returns an error that can be handled by the caller.

pkg/scheduler/scheduler.go
karlkfi: Fatal panics are idiomatic in "New" constructors. If we want to return an error we can wrap it in another static constructor that validates the config input and returns an error.

pkg/scheduler/uid/uid.go
karlkfi: good catch. is there a good way to indicate "unsigned"? do we need to?

pkg/scheduler/service/service.go
karlkfi: The current group would only be 0 if the current id is "", right? The 0 was coming from the default value of Group() from a nil (unparsable) UID.

karlkfi:
@jdef Should offers be rejected if the executor name doesn't match? It seems odd to reject if name matches and group doesn't without also rejecting when name doesn't match...

karlkfi:
For the record, I tried to keep the scope of this as small as possible. In the future we'll want to address the protobuff code leaking all over and try to hide it as an implementation detail, rather than having it permeate all the internal APIs. 

The primary motivation for this bit of work was to avoid having to re-parse or re-format the executor IDs multiple times.

pkg/scheduler/executor_ref.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"bytes"
+	"fmt"
+	"hash/crc64"
+	"sort"
+	"strconv"
+
+	"github.com/gogo/protobuf/proto"
+
+	execcfg "github.com/mesosphere/kubernetes-mesos/pkg/executor/config"
```
jdef: we don't need to alias this package here

pkg/scheduler/executor_ref.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (e *ExecutorRef) Command() *mesos.CommandInfo {
+	return e.proto.Command
+}
+
+func (e *ExecutorRef) Data() []byte {
+	return e.proto.Data
+}
+
+func (e *ExecutorRef) Proto() *mesos.ExecutorInfo {
```
jdef: I'd prefer a name other than `Proto` here, maybe `Info`?

pkg/scheduler/executor_ref.go
```diff
@@ -0,0 +1,126 @@
...
+		id: id,
+	}
+}
+
+func (e *ExecutorRef) ID() uid.UID {
+	return e.id
+}
+
+func (e *ExecutorRef) Name() string {
+	return *e.proto.Name
```
jdef: should probably be `e.proto.GetName()`

pkg/scheduler/executor_ref.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (e *ExecutorRef) Source() string {
+	return *e.proto.Source
```
jdef: should probably be `e.proto.GetSource()`

pkg/scheduler/executor_ref.go
```diff
@@ -0,0 +1,126 @@
...
+// errors at run-time because a Mesos master decides that the ExecutorInfo generated by a
+// secondary scheduler doesn't match that of the primary scheduler.
+//
+// see https://github.com/apache/mesos/blob/0.22.0/src/common/type_utils.cpp#L110
+func hashCode(info *mesos.ExecutorInfo) uint64 {
+	// !!! we specifically do NOT include:
+	// - Framework ID because it's a value that's initialized too late for us to use
+	// - Executor ID because it's a value that includes a copy of this hash
+	buf := &bytes.Buffer{}
+	buf.WriteString(*info.Name)
```
jdef: info.GetName()

pkg/scheduler/executor_ref.go
```diff
@@ -0,0 +1,126 @@
...
+	buf.WriteString(*info.Source)
```
jdef: info.GetSource()

pkg/scheduler/scheduler.go
```diff
@@ -199,6 +177,37 @@ func New(config Config) *KubernetesScheduler {
 	return k
 }
 
+// isCompatible returns true if the specified offer is compatible with the scheduler's executor.
+// Offers are compatible unless the name matches and the group doesn't.
+func (k *KubernetesScheduler) isOfferCompatible(o *mesos.Offer) bool {
+	for _, eid := range o.GetExecutorIds() {
+		executorIdString := eid.GetValue()
+		executorID, err := uid.Parse(executorIdString)
+		if err != nil {
```
jdef: offers are expected to arrive with executor IDs that are unrecognized by this framework. this logic would end up rejecting all offers from slaves that are running an executor whose ID is unparseable by us which is probably nearly all of them.

what we want to avoid the case where we validate an offer for resources on a slave that is running an executor launched by this framework (or some prior incarnation of it) with a different configuration (that's why we hash the executor info).

pkg/scheduler/service/service.go
```diff
@@ -492,17 +484,36 @@ func (s *SchedulerServer) awaitFailover(schedulerProcess schedulerProcessInterfa
...
-	d := uid.Parse(desired).Group()
-	c := uid.Parse(current).Group()
-	if d == 0 {
+	desiredUID, err := uid.Parse(desired)
+	if err != nil {
+		return fmt.Errorf("Failed to parse desired scheduler UID %q: %v", desired, err)
+	}
+
+	desiredGroup := desiredUID.Group()
+	if desiredGroup == 0 {
```
jdef: since you're already erroring out on parse errors, why another check for 0 here?

pkg/scheduler/service/service.go
```diff
@@ -492,17 +484,36 @@ func (s *SchedulerServer) awaitFailover(schedulerProcess schedulerProcessInterfa
...
+		return nil
 	}
+
+	currentUID, err := uid.Parse(current)
+	if err != nil {
+		return fmt.Errorf("Failed to parse current scheduler UID %q: %v", desired, err)
+	}
+
+	currentGroup := currentUID.Group()
+	if desiredGroup != currentGroup && currentGroup != 0 {
```
jdef: since you've already returned in current doesn't parse, why check for 0 here?

pkg/scheduler/service/service.go
```diff
@@ -549,11 +560,13 @@ func (s *SchedulerServer) bootstrap(hks hyperkube.Interface, sc *schedcfg.Config
 		log.Warningf("user-specified reconcile cooldown too small, defaulting to %v", s.ReconcileCooldown)
 	}
 
-	executor, eid, err := s.prepareExecutorInfo(hks)
+	executorCmd, err := s.prepareExecutorCommand(hks)
 	if err != nil {
-		log.Fatalf("misconfigured executor: %v", err)
+		log.Fatalf("misconfigured etcd: %v", err)
```
jdef: etcd?

pkg/scheduler/service/service_test.go
```diff
@@ -90,3 +90,25 @@ func Test_awaitFailoverDoneFailover(t *testing.T) {
...
+func Test_validateLeadershipTransition(t *testing.T) {
+	if err := validateLeadershipTransition("1_name-1", ""); err != nil {
+		t.Errorf("expected validation to pass when current id is empty")
+	}
+
+	if err := validateLeadershipTransition("1_name-1", "1_name-2"); err != nil {
+		t.Errorf("expected validation to pass when desired group equals current group")
+	}
+
+	if err := validateLeadershipTransition("0_name-1", ""); err == nil {
```
jdef: see comments in validation func re: 0 checks - they were originally intended as "unparseable" checks

pkg/scheduler/uid/uid.go
```diff
@@ -15,55 +16,49 @@ type UID struct {
 	ser   string
 }
 
-func New(group uint64, name string) *UID {
+func New(group uint64, name string) UID {
 	if name == "" {
-		name = uuid.New()
+		log.Fatalf("name must not be empty")
```
jdef: panic(fmt.Sprintf(...)) instead

jdef:
2nd review complete, please address comments

karlkfi:
very very obsolete...

PR #262, opened 2015-05-07
Add some more tests for DelayQueue and DelayFIFO

sttts:


sttts:
@jdef can you take a look?

pkg/queue/delay_test.go
```diff
@@ -112,6 +129,37 @@ func TestDQ_sanity_check(t *testing.T) {
...
+func TestDQ_Offer(t *testing.T) {
+	t.Parallel()
+	assert := assert.New(t)
+
+	dq := NewDelayQueue()
+	delay := time.Second
+
+	added := dq.Offer(&testjob{})
+	if added {
+		t.Fatalf("offered job without deadline added")
```
joerg84: Tiny nit: Could we change the log message to something like "DelayQueue should not accept job without deadline"

joerg84:
One tiny nit, otherwise lgtm.

jdef:
LGTM

sttts:
@joerg84 improved the error messages.

In fact, I much prefer the assert or expect style of assumption. Then one usually gets proper messages for free. Using the assert module for all new tests now.

joerg84:
@sttts Thanks, error messages look much better!

PR #264, opened 2015-05-07
Kube dns

jdef:
initial support for k8s cluster-level DNS resolution, meaning that kubernetes pods searching for <service>.<namespace>.<kubernetes> should be able to find such services using the `kube-dns` add-on from upstream which runs skydns, etcd, and kube2sky in pod and publishes it as a service in the k8s cluster.

TODO:
- [x] integrate build and deploy support for add-on
- [x] test whether pods can actually find services using this add-on, as we're configuring it
- [x] update k8s universe packaging with DNS_ and SCHEDULER_SERVICE_ADDRESS config items
  - see https://github.com/mesosphere/universe/pull/105


hack/images/dcos/bootstrap.sh
```diff
@@ -196,8 +167,57 @@ $apply_uids
...
+  touch kill && exit 0
+
+for i in $obj; do
+  /opt/kubectl create -f ${sandbox}/\$i
+done
+EOF
+
+  sed -i -e '$i test -f kill && exec s6-svc -d $(pwd) || exec \\' ${service_dir}/kube_dns/finish
+
+  # it's ok if etcd-server-depends is also monitoring etcd-server, the order of ops is this:
```
jdef: s/monitoring etcd-server/monitoring apiserver/

hack/images/dns/kube2sky/kube2sky.go
```diff
@@ -0,0 +1,292 @@
+/*
```
jdef: we don't really need this source code here, we can use what comes from upstream

jdef: well, upon further thought, either way the code is vendored. since it's not used in the main scheduler builds, it might as well be vendored here (that'll also make godep package updates easier).

hack/images/dcos/README.md
```diff
@@ -23,5 +22,42 @@
...
 - [ ] support secure (TLS) apiserver
+
+## Walkthrough w/ the guestbook examples
+
+Some prerequisites:
+
+- set up a DCOS cluster
+- install the `kubernetes` app using the DCOS CLI tooling
+  - `dcos package install kubernetes`
+- build yourself a kubectl, either using the Kubernetes or Kubernetes-Mesos build process
```
ConnorDoyle: In the future for DCOS, we could build a simple service that exposes a correctly-configured `kubectl` using something like [hterm](https://github.com/chromium/hterm).  Then we could distribute a more portable thin client CLI.  cc @jsancio for current plans or standardizing a way to do this.

hack/images/dns/kube2sky/kube2sky.go
ConnorDoyle: Having the code vendored in one place might make it easier to keep up-to-date.  Our goal is to un-vendor as much as possible, so If it's not too much pain, could we avoid bringing this copy into the source tree?

hack/images/dcos/README.md
```diff
@@ -3,7 +3,6 @@
 ## build and push the docker image
 :; make push TARGET=$HOME/bin/km
 ```
-
```
ConnorDoyle: Could you add a high-level summary here about SkyDNS, kube2sky, and how it all interacts (or doesn't) with Mesos-DNS?

hack/images/dns/kube2sky/kube2sky.go
jdef: Actually, it's going to make rebasing a little more error prone. Because godep is not the best tool. And we're going to have to rebase at least one more time as part of the upstream effort. I'd rather just stick this here for now and clean it up later since it'll disappear once we upstream.

ConnorDoyle: Ok, fair enough.

hack/images/dcos/README.md
jdef: done

ConnorDoyle:
Cool!  Care to squash the commits?  Other than that, let's :ship: it.

parikhyash:
Any idea when this PR will be merged?
Hoping to use kube-dns soon with my Mesos cluster.

jdef:
squashed to 4 commits, will merge once CI completes

sttts:
:+1:

PR #274, opened 2015-05-13
Add scheduler plugin tests

sttts:


jdef:
thanks for taking this on

pkg/scheduler/plugin.go
```diff
@@ -615,13 +615,12 @@ func (k *deleter) deleteOne(pod *Pod) error {
...
 
-// Create creates a scheduler plugin and all supporting background functions.
+// Create a scheduler plugin and all supporting background functions.
 func (k *KubernetesScheduler) NewPluginConfig(terminate <-chan struct{}, mux *http.ServeMux) *PluginConfig {
-
 	// Watch and queue pods that need scheduling.
 	updates := make(chan queue.Entry, k.schedcfg.UpdatesBacklog)
 	podUpdates := &podStoreAdapter{queue.NewHistorical(updates)}
-	reflector := cache.NewReflector(createAllPodsLW(k.client), &api.Pod{}, podUpdates, 0)
+	reflector := cache.NewReflector(k.podsListWatch, &api.Pod{}, podUpdates, 0)
```
jdef: instead of adding `podsListWatch` as a scheduler field, could we instead add an optional 3rd parameter to NewPluginConfig, `makePodsWatcher func() *cache.ListWatch`? Then, in NewPluginConfig:
```go
if makePodsWatcher == nil {
  makePodsWatcher = func() *cache.ListWatch { return createAllPodsLW(k.client) }
}
reflector := cache.NewReflector(makePodsWatcher(), &api.Pod{}, podUpdates, 0)
```
this allows us to:
- keep createAllPodsLW private
- avoids leaking podListWatch to the KubernetesScheduler struct
- keep setting up awesome unit/integration tests

sttts: Agree, that would be cleaner. Was happy to get that mocked without faking the whole apiserver watch API first.

jdef: I bet! It's tedious sometimes to figure out exactly how to get at mocking
things.

sttts: implemented.

sttts:
Ready for review @joerg84, @jdef

pkg/scheduler/plugin_test.go
```diff
@@ -1,15 +1,525 @@
...
+			},
+		},
+	}
+}
+
+func NewTestOffer(i int) *mesos.Offer {
+	hostname := fmt.Sprintf("h%d", i)
+	cpus := util.NewScalarResource("cpus", 3.75)
+	mem := util.NewScalarResource("mem", 940)
+	var port8000 uint64 = 8000
```
jdef: refactor to use podtask.Ranges once that's been merged

pkg/scheduler/plugin_test.go
```diff
@@ -1,15 +1,525 @@
...
+
+	// elect master with mock driver
+	driverFactory := ha.DriverFactory(func() (bindings.SchedulerDriver, error) {
+		return &mockDriver, nil
+	})
+	schedulerProcess.Elect(driverFactory)
+	elected := schedulerProcess.Elected()
+
+	// driver will be started
+	assert.EventuallyTrue(time.Second, func() bool { return len(mockDriver.CallsFor("Start")) > 0 })
```
jdef: :+1:

pkg/scheduler/plugin_test.go
```diff
@@ -1,15 +1,525 @@
...
+	pod = startPod(offers1)
+	failPodFromExecutor(launchTasks_taskInfos[0])
+
+	// 3. with pod still on the apiserver, bound i.e. host!=""
+	pod = startPod(offers1)
+	pod.Status.Host = *offers1[0].Hostname
+	podListWatch.Modify(pod, false) // not notifying the watchers
+	failPodFromExecutor(launchTasks_taskInfos[0])
+
+	// 3. with pod still on the apiserver, bound i.e. host!="", same
```
jdef: s/3./4./

jdef:
Really nice work. It occurs to me that it could be useful to reuse some of the test harness you've built here in other scheduler tests. We should keep that in mind as we add future tests.

Let me know when you're finished with the PR and I'll give it a final review.

sttts:
Ready for final review /cc @jdef 

As you say, there are a number of re-usable concepts. I plan to add more scheduler tests later (or we make a ticket out of them), then it makes sense to reuse those. When looking at the `scheduler.go` test coverage one sees a number of (especially non-standard) code paths which could be tested with that.

PR #278, opened 2015-05-14
DCOS release improvements

jdef:
- improvements to the dcos image release process
  - can still build & push -dev images easily for testing
  - can promote latest -dev build image to release-tagged via `promote` target
- bug fixes for service dependency mgmt

### DIFF
--- a/hack/images/dcos/Makefile
+++ b/hack/images/dcos/Makefile
@@ -48,7 +48,7 @@ DCOS_PREFIX ?= k8s-0.14.2-k8sm-0.5-dcos
 TAG	?= $(DCOS_PREFIX)-$(DCOS_VERSION)
 FQIN	?= $(REPO):$(TAG)
 
-.PHONY:	clean build push s6 $(S6_BINS) deps etcd $(ETCD_BINS) $(KUBE_DNS_TEMPLATES) version release promote
+.PHONY:	clean build push s6 $(S6_BINS) deps etcd $(ETCD_BINS) $(KUBE_DNS_TEMPLATES) version release promote check-sudo-docker info
 
 all:	build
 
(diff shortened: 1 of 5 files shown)


hack/images/dcos/Makefile
```diff
@@ -2,7 +2,7 @@ FROM	?= /target
 TARGET_OBJ = km kubectl
 IMAGE	?= kubernetes
 REPO	?= mesosphere/$(IMAGE)
-TAG	?= k8s-0.14.2-k8sm-0.5-dcos-dev
+DCOS_VERSION ?= dev
 SUDO	?= $(shell test "$$(whoami)" = "root" || echo sudo)
```
ConnorDoyle: Instead of calling `sudo` maybe we could make a target that can check `whoami` and fail with a descriptive message.  We'd depend on that target from the targets that need root.  I'd rather have to explicitly type `sudo make ...`, but maybe that's just me.

Probably doesn't matter much if the plan is to eventually scrap this Makefile.

Guess it doesn't matter too much if this is

hack/images/dcos/Makefile
```diff
@@ -27,7 +27,28 @@ KUBE_DNS_TEMPLATES_SOURCES = $(KUBE_DNS_TEMPLATES:%=../dns/%)
...
-.PHONY:	clean build push s6 $(S6_BINS) deps etcd $(ETCD_BINS) $(KUBE_DNS_TEMPLATES)
+#
+# normally you'll be generating dev images via:
+#	make release 
+#
+# and promoting (re-tagging them with a release version) them via:
+#	make promote
+#
+# HOWEVER, to generate a fresh release image and push it up (skip the -dev tag cycle):
+#	make release FROM=~/bin RELEASE=1
```
ConnorDoyle: This would be useful to note in the README as well.

hack/images/dcos/Makefile
jdef: Good feedback. Some thoughts:

- 'sudo make' scares me in general. I sure hope `rm -rf /` (or something
else destructive) never accidentally executes in this Makefile because I
hosed some variable.
- 'sudo docker ..do something..' fails and should abort the make process if
the user doesn't have sudo privs, or fat-fingers the password -- and
hopefully provides a useful error message.
- I like the idea of a privs-test target. /etc/sudoers config can be pretty
specific, so we'd probably have to test some docker noop like `sudo docker
ps >/dev/null` if we wanted (a slightly more reliable way) to validate
sudo-docker privs.
- not sure what the long term plan is for this Makefile. eventually we'll
probably deploy k8sm using marathon groups, but I haven't thought very far
ahead w/ respect to the evolution of this file -- working as I go. It'll
certainly want some cleanup if we decide to keep it around for the long
haul.

hack/images/dcos/Makefile
jdef: will do

jdef: done

hack/images/dcos/Makefile
jdef: added 'info' and 'check-sudo-docker' targets; targets that call docker now depend on `check-sudo-docker`

ConnorDoyle:
:shipit:

PR #281, opened 2015-05-18
Use gofmt -s to match k8s formatting requirements

### LINKED ISSUE #280: adopt k8s code formatting requirements
they're using `gofmt -s` now, which is different than what our `make format` target does. we need to be in compliance in order to upstream.

tsenart:
Fixes #280

### DIFF
--- a/pkg/scheduler/podtask/port_mapping.go
+++ b/pkg/scheduler/podtask/port_mapping.go
@@ -134,7 +134,7 @@ func defaultHostPortMapping(t *T, offer *mesos.Offer) ([]HostPortMapping, error)
 		}
 	}
 	foreachRange(offer, "ports", func(bp, ep uint64) {
-		for port, _ := range requiredPorts {
+		for port := range requiredPorts {
 			log.V(3).Infof("evaluating port range {%d:%d} %d", bp, ep, port)
 			if (bp <= port) && (port <= ep) {
 				mapping = append(mapping, requiredPorts[port])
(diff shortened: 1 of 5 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -68,7 +68,7 @@ func (self *slaveStorage) getSlaveIds() []string {
 	self.Lock()
 	defer self.Unlock()
 	slaveIds := make([]string, 0, len(self.slaves))
-	for slaveID, _ := range self.slaves {
+	for slaveID := range self.slaves {
```
tsenart: Was this code correct? It seems we were collecting the indexes, not the values in the slice.

pkg/scheduler/podtask/port_mapping.go
```diff
@@ -147,7 +147,7 @@ func defaultHostPortMapping(t *T, offer *mesos.Offer) ([]HostPortMapping, error)
 		err := &PortAllocationError{
 			PodId: t.Pod.Name,
 		}
-		for p, _ := range requiredPorts {
+		for p := range requiredPorts {
```
tsenart: As below, was this code correct?

pkg/scheduler/scheduler.go
sttts: Good catch.

tsenart: I'll follow up with a test and patch for these two cases.

tsenart: But I'll defer the impact analysis to @jdef.

jdef: `self.slaves` is a map, so `slaveID` should actually be an ID.

pkg/scheduler/podtask/port_mapping.go
jdef: `requiredPorts` is a map, so we're iterating over the keys.

jdef:
LGTM :ship:

PR #282, opened 2015-05-18
Add docker & docker-compose scripts

karlkfi:
This PR includes several pieces:
- A dockerfile for creating mesosphere/kubernetes-mesos-complete with all dependencies
- A dockerfile for creating mesosphere/kubernetes-mesos without dependencies
- A docker-compose file to run k8sm and each of its dependencies in separate containers
- Several scripts:
  - km-complete for running k8sm with dependencies
  - km-local for running k8sm without dependencies
  - helper scripts: await-service, cmd-exists, publicip, resolveip
- A readme for using the above

I figure these are useful on their own without the integration tests that prompted their creation. So I'm making the integration tests a different PR.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,113 @@
+# Running Kubernetes-Mesos In Docker
+
+There are several ways to play with kubernetes-mesos in docker. Here are a couple that might be most useful:
+
+1. [Complete](#complete)
+  - Run components and dependencies all in the same container
+  - All containers mirroring localhost ports (docker host mode)
+2. [Composed](#composed)
+  - Run components and dependencies each in different containers
+  - Each container gets its own IP
+  - Requires an ambassador to proxy ports (docker doesn't allow circular hostname linking)
+  - Emulates a more complex system, useful for testing networking
+
+While it is technically possible to run docker within a docker container, doing so requires running the mesos (slave) in priveledged mode.
+Instead, it's easier to mount the docker socket (`-v "/var/run/docker.sock:/var/run/docker.sock"`) and give mesos access to the docker it's running in.
+
+In addition to running in docker, it's also possible to [run docker locally](#local).
+
+<a name="complete"/>
+## K8SM Complete
+
+The "complete" Dockerfile includes everything needed to run a development instance of kubernetes-mesos, including etcd and mesos.
+
+### Daemon mode
+
+```
+docker run -d --name kubernetes-mesos -p 8888:8888 -p 5050:5050 -p 4001:4001 -v "/var/run/docker.sock:/var/run/docker.sock" mesosphere/kubernetes-mesos-complete
+```
+
+To see the logs:
+
+```
+docker logs kubernetes-mesos
+```
+
+To attach in interactive mode to a container already running in daemon mode:
+
+```
+docker exec -it kubernetes-mesos /bin/bash
+```
+
+### Interactive mode
+
+```
+docker run -it --name kubernetes-mesos -p 8888:8888 -p 5050:5050 -p 4001:4001 -v "/var/run/docker.sock:/var/run/docker.sock" --entrypoint=/bin/bash mesosphere/kubernetes-mesos-complete
+```
+
+Note: Interactive mode launches bash instead of the start script.
+
+### Stopping
+
+```
+docker kill kubernetes-mesos
+```
+
+### Building
+
+```
+./docker/complete/build.sh
+```
+
+
+<a name="composed"/>
+## K8SM Composed
+
+This method requires [Docker Compose](https://docs.docker.com/compose/) to be installed, which can be done via apt-get, homebrew, or [manually](https://docs.docker.com/compose/install/).
+
+The provided docker-compose.yml contains a self-contained configuration for running kubernetes-mesos, including its dependencies (etcd & mesos).
+It will launch 5 docker containers linked together with hostnames and port forwarding.
+
+```
+# from inside the docker dir
+docker-compose up
+```
+
+This will tail all the logs for each of the containers into STDOUT. `ctrl-c` to exit.
+
+### Stopping
+
+```
+docker-compose rm
+```
+
+### Building
+
+The docker-compose.yml file references multiple different docker images.
+
+The one that contains just kubernetes-mesos (not mesos or etcd) can be built with the following:
+
+```
+./docker/km/build.sh
+```
+
+
+<a name="local"/>
+## K8SM Local
+
+Since kubernetes-mesos requires etcd, mesos, and docker, these can all be started at once or individually.
+
+To start etcd, mesos, and kubernetes-mesos to run locally use the following command:
+
+```
+$ ./docker/bin/km-complete
+```
+
+If you already have etcd and mesos running locally use the following to start just the three kubernetes-mesos components:
+
+```
+$ ./docker/bin/km-local
+```
+
+Notes:
+- The mesos slave (which runs the kubernetes-mesos executor) needs docker to be installed (or configured via tcp). By default it talks to it using the socket `/var/run/docker.sock`.
--- a/docker/bin/install-etcd.sh
+++ b/docker/bin/install-etcd.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+ETCD_ARCHIVE_URL=${ETCD_ARCHIVE_URL:-https://github.com/coreos/etcd/releases/download/v0.4.6/etcd-v0.4.6-linux-amd64.tar.gz}
+ETCD_ARCHIVE=/tmp/$(basename ${ETCD_ARCHIVE_URL})
+
+echo "Downloading etcd (${ETCD_ARCHIVE_URL})..."
+wget -q $ETCD_ARCHIVE_URL -O $ETCD_ARCHIVE
+echo "Installing etcd (/usr/local/bin/etcd)..."
+mkdir -p /tmp/etcd
+tar xf $ETCD_ARCHIVE -C /tmp/etcd
+mv /tmp/etcd/*/etcd* /usr/local/bin/
+
+rm -f $ETCD_ARCHIVE
--- a/docker/bin/install-go.sh
+++ b/docker/bin/install-go.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -e
+
+GOPATH=${GOPATH:-~/go}
+GO_ARCHIVE_URL=${GO_ARCHIVE_URL:-https://storage.googleapis.com/golang/go1.4.2.linux-amd64.tar.gz}
+GO_ARCHIVE=/tmp/$(basename ${GO_ARCHIVE_URL})
+
+echo "Downloading go (${GO_ARCHIVE_URL})..."
+mkdir -p $(dirname $GOROOT)
+wget -q $GO_ARCHIVE_URL -O $GO_ARCHIVE
+echo "Installing go (${GOROOT})..."
+tar xf $GO_ARCHIVE -C $(dirname $GOROOT)
+
+if [ ! -d $TMPDIR ]; then
+    mkdir -p $TMPDIR
+fi
+
+rm -f $GO_ARCHIVE
--- a/docker/bin/km-complete
+++ b/docker/bin/km-complete
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+
+# Run etcd, mesos-local, km apiserver, km controller-manager, and km scheduler.
+# Exiting (or failure of) this script will cause the above sub-processes to be killed.
+# Usage: km-local
+# Mac support requires gtimeout & pcregrep (brew install coreutils && brew install pcre)
+# Inputs:
+# PUBLIC_IP - IP the etcd, mesos, and km components will listen on (unless overridden by MESOS_IP or ETCD_IP)
+# MESOS_IP - IP that mesos is accessible on
+# MESOS_PORT - Port that mesos-master is accessible on (default 5050)
+# ETCD_IP - IP that etcd is accessible on
+# ETCD_PORT - Port that etcd is accessible on (default 4001)
+# LOG_DIR - Directory to write logs to (default /tmp/k8sm-logs)
+
+set -e
+
+# add the current dir to PATH
+bin=$(cd $(dirname $0) && pwd -P)
+export PATH=$PATH:${bin}
+
+# add kubernetes-mesos/bin to PATH (unless km-local has been moved)
+k8sm_bin=$(cd ${bin}/../../bin && pwd -P)
+[ -d "$k8sm_bin" ] && [ -e "${k8sm_bin}/km" ] && export PATH=$PATH:${k8sm_bin} || true
+
+# validate required scripts & binaries
+cmd-exists "await-connection" "etcd" "mesos-local" "km-local" "publicip"
+
+PUBLIC_IP=$(publicip)
+
+set -e
+
+#TODO(karl): use mesos defaults & auto-detect ip/port from local mesos process
+export MESOS_IP=${PUBLIC_IP}
+export MESOS_PORT=5050
+export MESOS_MASTER=${MESOS_IP}:${MESOS_PORT}
+echo "Mesos: ${MESOS_MASTER}"
+
+#TODO(karl): use etcd defaults & auto-detect ip/port from local etcd process
+export ETCD_IP=${PUBLIC_IP}
+export ETCD_PORT=4001
+export ETCD_URL=http://${ETCD_IP}:${ETCD_PORT}
+echo "Etcd: ${ETCD_URL}"
+
+LOG_DIR=${LOG_DIR:-"/tmp/k8sm-logs"}
+mkdir -p ${LOG_DIR}
+echo "Log Dir: ${LOG_DIR}"
+
+echo "---------------------"
+
+
+source ${bin}/subprocs.sh
+
+# propagate SIGINT & SIGTERM
+for sig in INT TERM EXIT; do
+    trap "kill_subprocs ${sig}" ${sig}
+done
+
+
+echo "Starting etcd &> ${LOG_DIR}/etcd.log"
+etcd \
+  -addr=${ETCD_IP}:${ETCD_PORT} \
+  &> ${LOG_DIR}/etcd.log \
+  &
+add_subproc "$!" "etcd"
+#await-connection ${ETCD_IP} ${ETCD_PORT}
+await-health-check ${ETCD_URL}/health
+echo "---------------------"
+
+
+echo "Starting mesos-local ${LOG_DIR}/mesos.log"
+mesos-local \
+  &> ${LOG_DIR}/mesos.log \
+  &
+add_subproc "$!" "mesos-local"
+#await-connection ${MESOS_IP} ${MESOS_PORT}
+await-health-check http://${MESOS_MASTER}/health
+echo "---------------------"
+
+
+echo "Starting km-local"
+km-local &
+add_subproc "$!" "km-local"
+echo "---------------------"
+
+
+await_subprocs
--- a/docker/bin/km-local
+++ b/docker/bin/km-local
@@ -1,0 +1,126 @@
+#!/usr/bin/env bash
+
+# Run etcd, mesos-local, km apiserver, km controller-manager, and km scheduler.
+# Exiting (or failure of) this script will cause the above sub-processes to be killed.
+# Usage: km-local
+# Mac support requires gtimeout & pcregrep (brew install coreutils && brew install pcre)
+# Inputs:
+# PUBLIC_IP - IP the km components will listen on
+# MESOS_IP - IP that mesos is accessible on
+# MESOS_PORT - Port that mesos-master is accessible on (default 5050)
+# ETCD_IP - IP that etcd is accessible on
+# ETCD_PORT - Port that etcd is accessible on (default 4001)
+# LOG_DIR - Directory to write logs to (default /tmp/k8sm-logs)
+
+set -e
+
+# add the current dir to PATH
+bin=$(cd $(dirname $0) && pwd -P)
+export PATH=$PATH:${bin}
+
+# validate required scripts
+cmd-exists "await-connection"
+
+# add kubernetes-mesos/bin to PATH (unless km-local has been moved)
+k8sm_bin=$(cd ${bin}/../../bin && pwd -P)
+[ -d "$k8sm_bin" ] && [ -e "${k8sm_bin}/km" ] && export PATH=$PATH:${k8sm_bin} || true
+
+# validate required binaries
+cmd-exists "km"
+
+PUBLIC_IP=$(publicip)
+
+set -e
+
+export K8SM_IP=${PUBLIC_IP}
+export K8SM_PORT=8888
+echo "Kubernetes: ${K8SM_IP}:${K8SM_PORT}"
+
+#TODO(karl): auto-detect ip/port from local mesos process
+export MESOS_IP=${MESOS_IP:-$PUBLIC_IP}
+export MESOS_PORT=${MESOS_PORT:-5050}
+export MESOS_MASTER=${MESOS_IP}:${MESOS_PORT}
+echo "Mesos: ${MESOS_MASTER}"
+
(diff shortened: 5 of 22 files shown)


karlkfi:
Assigning @jdef 
CCing @ConnorDoyle & @joerg84 for secondary review.

docker/bin/await-service
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+# Wait for a service to accept connections.
+# Block up to the max wait time (default: 10 seconds).
+# Usage: timeout <ip> <port> [max_wait]
+
+set -e
+
```
spacejam: [ $# -ge 2 ] || usage

docker/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+<a name="complete"/>
+## K8SM Complete
+
+The "complete" Dockerfile includes everything needed to run a development instance of kubernetes-mesos, including etcd and mesos.
+
+### Background mode
+
+```
+docker run --name kubernetes-mesos -p 8888:8888 -p 5050:5050 -p 4001:4001 mesosphere/kubernetes-mesos &> /tmp/kubernetes-mesos-docker.log &
```
sttts: Instead of & use -d for daemon mode. Then you get the docker id.

docker/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+```
+docker kill kubernetes-mesos
+```
+
+### Building
+
+```
+# from root of project (not inside the docker dir)
+docker build -f ./docker/complete/Dockerfile -t mesosphere/kubernetes-mesos-complete .
```
sttts: above you use the image kubernetes-mesos, not k8sm-complete. Intentional?

docker/README.md
```diff
@@ -0,0 +1,105 @@
...
+The provided docker-compose.yml contains a self-contained configuration for running kubernetes-mesos, including its dependencies (etcd & mesos).
+It will launch 5 docker containers linked together with hostnames and port forwarding.
+
+```
+# from inside the docker dir
+docker-compose up
+```
+
+This will tail all the logs for each of the containers into STDOUT.
+`ctrl-C` to exit.
```
sttts: ctrl-c, no capital C

docker/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Building
+
+The docker-compose.yml file references multiple different docker images.
+
+The one that contains just kubernetes-mesos (not mesos or etcd) can be built with the following:
+
+```
+# from root of project (not inside the docker dir)
+docker build -f ./docker/complete/Dockerfile -t mesosphere/kubernetes-mesos-complete .
```
sttts: Why not merge that with the upper Building instructions?

docker/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+Since kubernetes-mesos requires etcd, mesos, and docker, these can all be started at once or individually.
+
+To start etcd, mesos, and kubernetes-mesos to run locally use the following command:
+
+```
+$ ./docker/bin/km-complete
+```
+
+If you already have etc and mesos running locally use the following to start just the three kubernetes-mesos components:
```
sttts: etcd

docker/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+```
+$ ./docker/bin/km-local
+```
+
+Notes:
+- The kubernetes executor, and thus the mesos slave, needs docker to be installed (or configured via tcp). By default it talks to it using the socket `/var/run/docker.sock`.
```
sttts: More like the other way around, the mesos-slave needs docker (the containerizer actually) and therefore the k8sm executor as well which makes use of the containerizer of the slave.

docker/complete/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+EXPECTED_SCRIPT_DIR=/docker/complete
+
+set -e
+
+script_dir=$(cd $(dirname $0) && pwd -P)
+
+# desense against lazy refactoring
```
sttts: defence

docker/README.md
sttts: Usually one leaves out the &> ... redirection and just uses `docker logs` afterwards.

docker/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
...
+apiserver:
+  hostname: apiserver
+  image: mesosphere/kubernetes-mesos:latest
+  entrypoint:
+  - /bin/bash
+  - "-c"
+  - >
+    echo "Hostname: $(hostname -f) ($(hostname -f | xargs resolveip))" &&
+    echo "mesos-master = mesos:5050" >> /opt/mesos-cloud.conf &&
+    await-service $(resolveip etcd) 4001 &&
```
sttts: A pity that also docker compose has nothing to wait for other services to come up. This is really a shortcoming of the whole Docker approach.

docker/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
...
+#  - zookeeper
+apiserver:
+  hostname: apiserver
+  image: mesosphere/kubernetes-mesos:latest
+  entrypoint:
+  - /bin/bash
+  - "-c"
+  - >
+    echo "Hostname: $(hostname -f) ($(hostname -f | xargs resolveip))" &&
+    echo "mesos-master = mesos:5050" >> /opt/mesos-cloud.conf &&
```
sttts: Doesn't this break when a container is started again (without rm'ing)?

docker/complete/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+        && \
+    apt-get clean
+
+COPY . /kubernetes-mesos
+WORKDIR /kubernetes-mesos
+
+ENV GOROOT /usr/local/go
+ENV PATH $GOROOT/bin:$PATH
+
+RUN ./docker/complete/install-go.sh
```
sttts: Docker uses pretty simple change detection while building. If the COPY above "changes something", the install scripts will be called again, including the download. It's much fast to put these two install lines above the local COPY.

docker/mesos-cloud.conf
```diff
@@ -0,0 +1,4 @@
+[mesos-cloud]
+  http-client-timeout = 5s
+  state-cache-ttl     = 20s
+
```
sttts: Without having the complete overview, can't you put the mesos:5050 line here as well? It's constant after all.

docker/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
...
+    --cloud_config=/opt/mesos-cloud.conf
+    --v=2
+  links:
+  - mesos
+  - apiserver
+scheduler:
+  hostname: scheduler
+  image: mesosphere/kubernetes-mesos:latest
+  entrypoint:
+  - /bin/bash
```
sttts: The entrypoint in string format is always run inside a shell by Docker. I guess there is no need for another bash line here. Completely sure yet, especially about the way docker-compose handles this.

docker/docker-compose.yml
karlkfi: Yes. The container must be rm'd. Do you know of a way to get around that?

sttts: Either put the line statically into the medos-cloud.conf or grep for it:

```
grep 'mesos-master ....' || echo .... >> ...
```

docker/mesos-cloud.conf
karlkfi: Yeah, there's some overlap between what km-local & docker-compose.yml do in regards to that config. If you run km-local outside of docker-compose you'll need a different IP in there.

docker/README.md
```diff
@@ -0,0 +1,113 @@
...
+  - Run components and dependencies all in the same container
+  - All containers mirroring localhost ports (docker host mode)
+2. [Composed](#composed)
+  - Run components and dependencies each in different containers
+  - Each container gets its own IP
+  - Requires an ambassador to proxy ports (docker doesn't allow circular hostname linking)
+  - Emulates a more complex system, useful for testing networking
+
+While it is technically possible to run docker within a docker container, doing so requires running the mesos (slave) in priveledged mode.
+So the simplest way is to instead expose the bind mount the docker socket (`-v "/var/run/docker.sock:/var/run/docker.sock"`)
```
jdef: grammar

docker/bin/await-service
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+# Wait for a service to accept connections.
```
jdef: I like the idea behind this script. Keep in mind that it can lie about etcd. Etcd will open a socket listener before it's actually ready to process API requests. In the DCOS image I've build, I ended up waiting on an HTTP 200 from some service URL instead of host:port.

docker/bin/install-etcd.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -ex
+
+ETCD_ARCHIVE_URL=https://github.com/coreos/etcd/releases/download/v0.4.6/etcd-v0.4.6-linux-amd64.tar.gz
```
jdef: might be nice to rewrite as defaulting to an external source, so that we could more easily test different etcd releases:
```sh
ETCD_ARCHIVE_URL=${ETCD_ARCHIVE_URL:-https://github.com/coreos/etcd/releases/download/v0.4.6/etcd-v0.4.6-linux-amd64.tar.gz}
```

docker/bin/install-go.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -ex
+
+GOPATH=${GOPATH:-~/go}
+GO_ARCHIVE_URL=https://storage.googleapis.com/golang/go1.4.2.linux-amd64.tar.gz
```
jdef: same comment here as for ETCD_ARCHIVE_URL

docker/bin/km-complete
```diff
@@ -0,0 +1,140 @@
...
+# add the current dir to PATH
+bin=$(cd $(dirname $0) && pwd -P)
+export PATH=$PATH:${bin}
+
+# add kubernetes-mesos/bin to PATH (unless km-local has been moved)
+k8sm_bin=$(cd ${bin}/../../bin && pwd -P)
+[ -d "$k8sm_bin" ] && [ -e "${k8sm_bin}/km" ] && export PATH=$PATH:${k8sm_bin} || true
+
+# validate required scripts & binaries
+cmd-exists "await-service"
```
jdef: what about writing cmd-exists to accept multiple cmd args?
```sh
cmd-exists etcd mesos-local km-local ...
```

docker/bin/km-complete
```diff
@@ -0,0 +1,140 @@
...
+                kill ${SUBPROC} || true
+            fi
+        done
+    fi
+    if [ "${SIGNAL}" != "EXIT" ]; then
+        trap EXIT
+    fi
+}
+
+# propergate SIGINT & SIGTERM
```
jdef: s/propergate/propagate

docker/bin/km-complete
```diff
@@ -0,0 +1,140 @@
...
+    SIGNAL=$1
+    echo "Shutting down [km-complete] (Signal: ${SIGNAL})"
+    if [ ! -z "${SUBPROCS}" ]; then
+        for ((i=${#SUBPROCS[@]}-1; i>=0; i--)); do
+            local SUBPROC=${SUBPROCS[i]}
+            local SUBPROC_NAME=${SUBPROC_NAMES[i]}
+            local SUBPROC_NAME=${SUBPROC_NAME//_/ } # replace underscores with spaces
+            if pid_running "${SUBPROC}"; then
+                echo "Killing ${SUBPROC_NAME} (${SUBPROC})"
+                #TODO(karl): propegate the recieved signal, instead of just TERM
```
jdef: s/propegate/propagate

docker/bin/km-complete
```diff
@@ -0,0 +1,140 @@
...
+# propergate SIGINT & SIGTERM
+for sig in INT TERM EXIT; do
+    trap "kill_subprocs ${sig}" ${sig}
+done
+
+
+echo "Starting etcd &> ${LOG_DIR}/etcd.log"
+etcd \
+  -addr=${ETCD_IP}:${ETCD_PORT} \
+  &> ${LOG_DIR}/etcd.log \
```
jdef: you may want to consider log file rotation at some point

docker/bin/km-local
```diff
@@ -0,0 +1,174 @@
...
+
+PUBLIC_IP=$(publicip)
+
+set -e
+
+export K8SM_IP=${PUBLIC_IP}
+export K8SM_PORT=8888
+echo "Kubernetes: ${K8SM_IP}:${K8SM_PORT}"
+
+export MESOS_IP=${MESOS_IP:-$PUBLIC_IP}
```
jdef: these env declarations look pretty similar to those in `km-complete`. opportunity for refactoring into some `common.sh` ?

docker/docker-compose.yml
karlkfi: I can add something like that, but unfortunately the etcd container already doesn't support restarting without removal. It croaks with `System error: exec: "/bin/echo": stat /bin/echo: no such file or directory`

docker/bin/km-local
```diff
@@ -0,0 +1,174 @@
...
+SUBPROCS=()
+SUBPROC_NAMES=()
+
+pid_running () {
+    local CMD_PID=$1
+    return $(kill -0 ${CMD_PID} &> /dev/null)
+}
+
+# Add the pid of a sub-process or exit if it is not running
+add_subproc () {
```
jdef: similar comment re: refactoring opportunity

docker/bin/km-local
```diff
@@ -0,0 +1,174 @@
...
+                kill ${SUBPROC} || true
+            fi
+        done
+    fi
+    if [ "${SIGNAL}" != "EXIT" ]; then
+        trap EXIT
+    fi
+}
+
+# propergate SIGINT & SIGTERM
```
jdef: propagate

docker/bin/km-local
```diff
@@ -0,0 +1,174 @@
...
+
+echo "Starting km apiserver &> ${LOG_DIR}/apiserver.log"
+km apiserver \
+  --address=${K8SM_IP} \
+  --etcd_servers=${ETCD_URL} \
+  --portal_net=10.10.10.0/24 \
+  --port=${K8SM_PORT} \
+  --cloud_provider=mesos \
+  --cloud_config=${K8SM_CONFIG} \
+  &> ${LOG_DIR}/apiserver.log \
```
jdef: may want to think about log file rotation

docker/bin/publicip
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ -z "${PUBLIC_IP}" ]; then
+    if cmd-exists "ifconfig" && cmd-exists "ipconfig" && cmd-exists "pcregrep"; then
+        PUBLIC_IP=$(ifconfig | pcregrep -M -o '^[^\t:]+:([^\n]|\n\t)*status: active' | egrep -o -m 1 '^[^\t:]+' | xargs ipconfig getifaddr)
+    fi
+    if [ -z "${PUBLIC_IP}" ] && cmd-exists "ip"; then
+        PUBLIC_IP=$(ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1  -d'/')
```
jdef: may yield ipv6 addresses. may want to consider something like this:
```sh
ip -o -f inet addr | grep -v -e 'inet 127\.' | head -1 | sed -e 's,^.*inet \([^/]\+\)/.*$,\1,g'
```
or, grab the IP of the interface of the default route:
```sh
defaultdev=$(ip route|grep ^default|tr ' ' '\n'|sed -e '/^$/d'|tail -1)
ip -o -f inet addr show $(defaultdev) | grep -v -e 'inet 127\.' | head -1 | sed -e 's,^.*inet \([^/]\+\)/.*$,\1,g'
```
still haven't finished a complete review of the PR, so maybe this is overkill if publicip only runs in-container

docker/complete/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM mesosphere/mesos:0.22.0-1.0.ubuntu1404
+MAINTAINER Mesosphere <support@mesosphere.io>
+
+RUN locale-gen en_US.UTF-8
+RUN dpkg-reconfigure locales
+ENV LANG en_US.UTF-8
+ENV LC_ALL en_US.UTF-8
+
```
jdef: may want
```
ENV DEBIAN_FRONTEND noninteractive
```

docker/complete/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN apt-get update && \
+    apt-get upgrade -y && \
```
jdef: why the `upgrade`? seems (a) not necessary, and; (b) to change from a known state/baseline to a moving target

docker/complete/build.sh
```diff
@@ -0,0 +1,49 @@
...
+cleanup() {
+    rm -rf ${WORKSPACE}
+    echo "Workspace deleted: $WORKSPACE"
+}
+trap 'cleanup' EXIT
+
+mkdir ${WORKSPACE}/bin
+
+echo "Building kubernetes-mesos binaries"
+docker run --rm -v ${WORKSPACE}/bin:/target mesosphere/kubernetes-mesos:build
```
jdef: this will always do a fresh `git clone`. you should be able to avoid that (if you want) by adding `-v /path/to/already-checked-out-repo:/snapshot` to the docker run arg list.

docker/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
...
+  - MESOS_HOSTNAME=mesos
+  - MESOS_PORT=5050
+  - MESOS_LOG_DIR=/var/log/mesos
+  - MESOS_QUORUM=1
+  - MESOS_REGISTRY=in_memory
+  - MESOS_WORK_DIR=/var/lib/mesos
+#  - MESOS_ZK=zk://zookeeper:2181/mesos
+  links:
+  - etcd
+  - "ambassador:apiserver
```
jdef: missing end " ?

docker/docker-compose.yml
jdef: waiting for other services to start is a crutch that should not be encouraged. Seems like a reasonable exclusion to me :)

docker/km/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:14.04.2
+MAINTAINER Mesosphere <support@mesosphere.io>
+
+RUN locale-gen en_US.UTF-8
+RUN dpkg-reconfigure locales
+ENV LANG en_US.UTF-8
+ENV LC_ALL en_US.UTF-8
+
```
jdef: ENV DEBIAN_FRONTEND noninteractive

docker/km/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:14.04.2
+MAINTAINER Mesosphere <support@mesosphere.io>
+
```
jdef: ENV DEBIAN_FRONTEND noninteractive

docker/complete/Dockerfile
jdef: .. before RUN'ing locale-gen or dpkg-reconfigure

pkg/scheduler/service/service.go
```diff
@@ -596,26 +594,29 @@ func (s *SchedulerServer) bootstrap(hks hyperkube.Interface, sc *schedcfg.Config
 	runtime.On(mesosPodScheduler.Registration(), func() { kpl.Run(schedulerProcess.Terminal()) })
 	runtime.On(mesosPodScheduler.Registration(), s.newServiceWriter(schedulerProcess.Terminal()))
 
-	deferredInit := func() (err error) {
+	driverFactory := ha.DriverFactory(func() (bindings.SchedulerDriver, error) {
```
jdef: thanks for the cleanup here.

jdef:
looks pretty nice overall. some doc nits and other small-ish things. may want to punt on log rotation for now (can file a separate issue to track that if we really want it).

docker/bin/await-service
karlkfi: Good point. I see Etcd has a /health endpoint for that purpose.
Do the other components have one we should use too?

jdef: apiserver has /healthz

there's a PR to add health endpoints to other apps. needs review.

docker/complete/Dockerfile
karlkfi: mesos container is on ubuntu, not debian. Is it the same on both?

docker/complete/Dockerfile
karlkfi: good point.

docker/docker-compose.yml
karlkfi: docker compose doesn't do any differently than docker, and docker uses "/bin/sh -c" by default. The commands might work in shell, but it's just less headache to use bash.

docker/bin/km-complete
karlkfi: good call.

docker/bin/km-local
karlkfi: I'm extracting the common functions to a `subprocs.sh` and hiding that timeout mess in `timeout`, but I'm not sure I want to extract the IPs, namely because they choose default values which I kinda hate doing in shell scripts. If there were a way to not have to hardcode defaults at all but instead pull the values from the running commands and use the command's defaults when launching them... that would be far superior.

docker/bin/km-local
karlkfi: Log rotation might be necessary, depending on usage, but I suspect deleting after X time or over Y size is the more important part. We can definitely talk about this, but it's not quite critical yet. I don't expect to need log rotation (other than maybe truncation) during a test run.

docker/complete/Dockerfile
karlkfi: Presuming your goal is to stop the stderr red text from showing up when docker is building, I'm adding `DEBIAN_FRONTEND=noninteractive` to JUST the apt-get command.

Based on the below issue, it seems that noninteractive would be wrong if the container is to be used in interactive mode, but is correct for the non-interactive image building.
https://github.com/docker/docker/issues/4032

Unfortunately the UNSET dockerfile command is still in limbo. So we can't just set it at the begining and unset it at the end.
https://github.com/docker/docker/pull/8177

karlkfi:
For future discussion:
- Add support for log rotation within the containers (or when using km-local or km-complete locally)
- Decide whether to always clone master or mount the local workspace (`-v ${project_dir}:/snapshot`) when building
- Does `docker/bin/publicip` need to support ipv6? https://github.com/mesosphere/kubernetes-mesos/pull/282#discussion_r30544543
- Find a way to avoid hard coding default IPs (0.0.0.0) & ports in shell scripts. Maybe use different/better k8sm defaults and/or auto-detect ip/port or running process/container?

karlkfi:
For what it's worth, the dockerhub etcd image is ANCIENT. Use quay.io/coreos/etcd instead.

docker/complete/Dockerfile
jdef: My goal is to avoid packages trying to ask the user questions. See http://www.microhowto.info/howto/perform_an_unattended_installation_of_a_debian_package.html

jdef:
nice job. like the changes :+1:

jdef:
@karlkfi Can we add issues for each of the below items (that link to this PR)? I don't want to hold up this PR and it's in good enough shape to start using for testing. But I don't want to lose track of these items either.

> For future discussion:
>
> * Add support for log rotation within the containers (or when using km-local or km-complete locally)
> * Decide whether to always clone master or mount the local workspace (-v ${project_dir}:/snapshot) when building
> * Does docker/bin/publicip need to support ipv6? #282 (comment)
> * Find a way to avoid hard coding default IPs (0.0.0.0) & ports in shell scripts. Maybe use different/better k8sm defaults and/or auto-detect ip/port or running process/container?

karlkfi:
I'm not really sure how to make my ports comment into an issue that makes sense. So I'm just going to drop it until/unless we feel more pain/confusion in that area.

jdef:
LGTM, ready for merge after squash

jdef:
:ship:

PR #284, opened 2015-05-19
sync arch diagram w/ current implementation, as of k8s v0.14 integration

jdef:
Updated diagrams, needed for related blueprint docs.


ConnorDoyle:
For the purposes of this diagram, I think we can drop the communication line between the slave process and dockerd.

ConnorDoyle:
Could we add a line between Zookeeper and the K-M scheduler (for master detection?)

ConnorDoyle:
When we do the next rebase we'll be able to pass the parent cgroup to the kubelet to make the executor container enclose the pods!

ConnorDoyle:
Where does the scheduler talk directly to etcd?

jdef:
scheduler stores frameworkID in etcd (pkg/scheduler/service). Also uses
etcd for ha leader election (pkg/election)

jdef:
Yep, super excited about that!

I think it'll save time to just upstream what we have first (vs. a separate
rebase cycle), and as part of that effort we can have the executor
container enclose the pods. We're getting close to 1.0 upstream code freeze
and I want to have a working upstream PR in place by then.

jdef:
@ConnorDoyle Thanks for the feedback, I've addressed your concerns. PTAL

ConnorDoyle:
LGTM!  :ship: :ship: :ship:

Issue #272, opened 2015-05-12
executor port is not allocated from mesos offer


jdef:
the executor uses at least three ports:

# executor driver port
# API port (kubelet/server.go)
# cadvisor port

the driver port can be dynamic, mesos doesn't care. the API and cadvisor ports should probably be the same across all executors in the cluster. none of these ports are currently allocated from resource offers.

sttts:
We don't set the driver port explicitly. Will Mesos and the mesos-go binding take care of registering the driver port correctly?

sttts:
There is a global resource mechanism in the work in Mesos. Maybe this would be the right way to get those cluster global ports:  https://issues.apache.org/jira/browse/MESOS-2742

jdef:
there's a DriverConfig.BindingPort that you can use

PR #286, opened 2015-05-19
Improve port mapping code

tsenart:
This change set improves port mapping code in preparation for #272. Test coverage should increase too.
The diff is really messy. Sorry for that... :/


pkg/scheduler/podtask/ranges.go
```diff
@@ -0,0 +1,140 @@
...
+type Range [2]uint64
+
+// Ranges represents a list of Ranges.
+type Ranges []Range
+
+// NewPortRanges returns Ranges from the "ports" resource in the
+// given *mesos.Offer. If that resource isn't provided, nil will be returned.
+//
+// The returned Ranges are sorted and have all overlapping ranges merged from
+// left to right. e.g. [[0, 5], [4, 3], [10, 7]] -> [[0, 5], [7, 10]]
```
spacejam: This seems like something I would expect to have already occurred.  Should this logic instead exist in our go mesos library and happen by default?

tsenart: The documentation on the `mesos.proto` definition says:
```
/**
 * Describes a resource on a machine. A resource can take on one of
 * three types: scalar (double), a list of finite and discrete ranges
 * (e.g., [1-10, 20-30]), or a set of items. A resource is described
 * using the standard protocol buffer "union" trick.
 *
 * TODO(benh): Add better support for "expected" resources (e.g.,
 * cpus, memory, disk, network).
 */
```

Unfortunately this doesn't really guarantee if the ranges are sorted or squashed correctly. The **discreteness** alluded to could have no gaps (e.g. `[[1, 10], [11,20]]`) which should be squashed (`[[1, 20]]`).

I wend down this road to for the sake of robustness. We should try to avoid relying on non-documented behaviour or implementation details.

Now, I totally agree this could be part of `mesos-go`! Good call. Though, I'm not sure if we should spend time on that right now. Perhaps we can keep it in our roadmap.

tsenart: Meanwhile, we could also ask some members of the core team if they can provide all of these guarantees in the documentation.

pkg/scheduler/podtask/ranges_test.go
```diff
@@ -0,0 +1,142 @@
...
+		{Ranges{}, Ranges{}},
+		{Ranges{{0, 1}}, Ranges{{0, 1}}},
+		{Ranges{{0, 2}, {1, 5}, {2, 10}}, Ranges{{0, 10}}},
+		{Ranges{{0, 2}, {2, 5}, {5, 10}}, Ranges{{0, 10}}},
+		{Ranges{{0, 2}, {3, 5}, {6, 10}}, Ranges{{0, 10}}},
+		{Ranges{{0, 2}, {4, 11}, {6, 10}}, Ranges{{0, 2}, {4, 11}}},
+		{Ranges{{0, 2}, {4, 5}, {6, 7}, {8, 10}}, Ranges{{0, 2}, {4, 10}}},
+		{Ranges{{0, 2}, {4, 6}, {8, 10}}, Ranges{{0, 2}, {4, 6}, {8, 10}}},
+		{Ranges{{0, 1}, {2, 5}, {4, 8}}, Ranges{{0, 8}}},
+	} {
```
tsenart: @spacejam: Take a look at the tests.

pkg/scheduler/plugin.go
```diff
@@ -188,17 +188,18 @@ func (b *binder) prepareTaskForLaunch(ctx api.Context, machine string, task *pod
...
-	for _, entry := range task.Spec.PortMap {
-		oemPorts := pod.Spec.Containers[entry.ContainerIdx].Ports
-		ports := append([]api.ContainerPort{}, oemPorts...)
-		p := &ports[entry.PortIdx]
-		p.HostPort = int(entry.OfferPort)
-		op := strconv.FormatUint(entry.OfferPort, 10)
-		pod.Annotations[fmt.Sprintf(annotation.PortMappingKeyFormat, p.Protocol, p.ContainerPort)] = op
-		if p.Name != "" {
-			pod.Annotations[fmt.Sprintf(annotation.PortNameMappingKeyFormat, p.Protocol, p.Name)] = op
+	for _, m := range task.Spec.PortMap {
```
jdef: the prior code made efforts to avoid modifying the task's copy of the pod spec. the new version does not.
why was this block changed? was it buggy?

pkg/scheduler/plugin.go
```diff
@@ -284,7 +285,7 @@ func (k *kubeScheduler) doSchedule(task *podtask.T, err error) (string, error) {
 		} else {
 			task.Offer.Release()
 			task.Reset()
-			if err = k.api.tasks().Update(task); err != nil {
+			if err := k.api.tasks().Update(task); err != nil {
```
jdef: nice catch

pkg/scheduler/podtask/port_mapping.go
```diff
@@ -2,168 +2,148 @@ package podtask
...
-type HostPortMapping struct {
-	ContainerIdx int // index of the container in the pod spec
-	PortIdx      int // index of the port in a container's port spec
-	OfferPort    uint64
+// NewPortMapper returns an appropriate PortMapper for a given Pod.
+func NewPortMapper(pod *api.Pod) PortMapper {
+	filter := map[string]string{PortMappingLabelKey: "fixed"}
+	selector := labels.Set(filter).AsSelector()
+	if selector.Matches(labels.Set(pod.Labels)) {
+		return "fixed"
```
jdef: "wildcard" and "fixed" were const before, why make them free-form?

pkg/scheduler/podtask/port_mapping.go
```diff
@@ -2,168 +2,148 @@ package podtask
...
-	return defaultHostPortMapping(t, offer)
 }
 
-type PortAllocationError struct {
-	PodId string
-	Ports []uint64
+// PortMapping represents a mapping from container ports to host ports for a
+// given pod container.
+type PortMapping struct {
+	PodName        string
```
jdef: why is PodName tracked here?

pkg/scheduler/podtask/port_mapping.go
```diff
@@ -2,168 +2,148 @@ package podtask
...
 }
 
-func (err *DuplicateHostPortError) Error() string {
-	return fmt.Sprintf(
-		"Host port %d is specified for container %d, pod %d and container %d, pod %d",
-		err.m1.OfferPort, err.m1.ContainerIdx, err.m1.PortIdx, err.m2.ContainerIdx, err.m2.PortIdx)
-}
+// byHostPort is an auxiliary type used to sort []PortMappings by their
+// HostPorts.
+type byHostPort []PortMapping
```
jdef: Like this pattern :+1:

pkg/scheduler/podtask/port_mapping.go
```diff
@@ -2,168 +2,148 @@ package podtask
...
-			}
+	// higher ports have highest priority, wildcard has the lowest
+	sort.Sort(sort.Reverse(byHostPort(wanted)))
+
+	taken := make(map[uint64]int, len(wanted))
+	for i := range wanted {
+		if offered.Size() < uint64(len(wanted)-len(taken)) {
+			break // no more available ports in the offer
+		} else if wanted[i].HostPort == 0 {
+			wanted[i].HostPort = offered.Min()
```
jdef: Don't you want a Pop() or Take() func instead of Min()? What if there are multiple 0 host ports?

jdef: nevermind. misread the if/else blocks

pkg/scheduler/podtask/port_mapping.go
```diff
@@ -2,168 +2,148 @@ package podtask
...
-				remaining--
-				taken[port] = struct{}{}
-				break
-			}
+	// higher ports have highest priority, wildcard has the lowest
+	sort.Sort(sort.Reverse(byHostPort(wanted)))
+
+	taken := make(map[uint64]int, len(wanted))
+	for i := range wanted {
+		if offered.Size() < uint64(len(wanted)-len(taken)) {
```
jdef: instead of referencing wanted[i].HostPort everywhere, wouldn't it be more readable to say:
```go
hostPort := &wanted[i].HostPort
```
and then reference that?

pkg/scheduler/podtask/port_mapping_test.go
```diff
@@ -1,189 +1,65 @@
...
-	mt = MappingTypeForPod(pod)
-	if mt != HostPortMappingFixed {
-		t.Fatalf("expected fixed mapping")
-	}
-
-	pod.Labels[PortMappingLabelKey] = string(HostPortMappingWildcard)
-	mt = MappingTypeForPod(pod)
-	if mt != HostPortMappingWildcard {
-		t.Fatalf("expected wildcard mapping")
+func container(hostports ...int) api.Container {
```
jdef: suggest `fakeContainer` (as opposed to `newContainer` or `container`); this is the naming convention that appears to be used upstream

pkg/scheduler/podtask/ranges.go
```diff
@@ -0,0 +1,140 @@
...
+		case 1+top[1] <= rs[i][1]: // overlap or continuity: squash
+			squashed[len(squashed)-1][1] = rs[i][1]
+		}
+	}
+	return squashed
+}
+
+// Find performs a binary search for n returning the index of the Range it was
+// found at or -1 if not found.
+func (rs Ranges) Find(n uint64) int {
```
jdef: I misread `Find` at first because I thought it was finding a port, not a port range. Could we rename to `RangeFor` for `RangeIndexOf`, or something to indicate that it's returning a range index?

pkg/scheduler/podtask/ranges.go
```diff
@@ -0,0 +1,140 @@
...
+			Begin: proto.Uint64(rs[i][0]),
+			End:   proto.Uint64(rs[i][1]),
+		}
+	}
+	return &mesos.Resource{
+		Name:   proto.String(name),
+		Type:   mesos.Value_RANGES.Enum(),
+		Ranges: &mesos.Value_Ranges{Range: vr},
+	}
+}
```
jdef: this ranges code is super-clean. nice work

pkg/scheduler/podtask/ranges_test.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+func TestNewPortRanges(t *testing.T) {
+	t.Parallel()
+
+	for i, tt := range []struct {
+		Ranges
+		want Ranges
+	}{
+		{Ranges{{2, 0}, {3, 10}}, Ranges{{0, 10}}},
```
jdef: what about a range that needs squashing, for completeness?

pkg/scheduler/plugin.go
tsenart: It was modified to be more readable. Why were you avoiding modification of the pod spec? Is being concurrently accessed? We have to update the containers anyways.

pkg/scheduler/podtask/port_mapping.go
tsenart: Refactoring artifact. I'll add the constants back.

pkg/scheduler/podtask/port_mapping.go
tsenart: For better error messages. See https://github.com/mesosphere/kubernetes-mesos/blob/refactor-port-mapping/pkg/scheduler/podtask/port_mapping.go#L121

pkg/scheduler/podtask/port_mapping.go
tsenart: I can use a copy of `wanted[i]` after the mutation in line 98. That'll make it more readable in the chunk.

pkg/scheduler/podtask/port_mapping_test.go
tsenart: Hmm, that will make the line width explode in the tests. What about just adding a comment to the function? Also, it's not really returning a fake container, it's returning a real container struct, since this isn't a mock implementation of an interface.

pkg/scheduler/podtask/ranges.go
tsenart: How could I improve the documentation of the method to make it more clear? The data type contains ranges, not individual numbers, so prepending `Range` to the method name seems superfluous.

pkg/scheduler/podtask/ranges_test.go
tsenart: The first case needs squashing due to overlap. I can add a case for squashing due to continuity.

tsenart:
@jdef: Please review last 4 commits.

pkg/scheduler/plugin.go
jdef: No, but the thinking was that we should be able to roll back if needed. If you overwrite the original values then that's not possible.

pkg/scheduler/podtask/port_mapping.go
jdef: ok, then can we make it package-private?

pkg/scheduler/podtask/ranges.go
jdef: The docs are great. I wish the func name was more descriptive. `Find` is pretty terse. even `IndexOf` would be an improvement.

pkg/scheduler/podtask/ranges_test.go
jdef: I think I'm missing something then. It looks to me like there are two adjacent ranges being squashed, not overlapping ranges.

pkg/scheduler/podtask/port_mapping.go
```diff
@@ -2,168 +2,154 @@ package podtask
...
 
-func (self HostPortMappingType) Generate(t *T, offer *mesos.Offer) ([]HostPortMapping, error) {
-	switch self {
-	case HostPortMappingWildcard:
-		return wildcardHostPortMapping(t, offer)
-	case HostPortMappingFixed:
+// PortMap computes PortMappings for the given Containers with the given offered port Ranges.
+func (m PortMapper) PortMap(offered Ranges, pod string, cs ...api.Container) ([]PortMapping, error) {
+	switch m {
+	case "wildcard":
```
jdef: I left a comment here and I don't see it now. So I'm leaving another one.
Should reference the consts here

pkg/scheduler/podtask/port_mapping.go
jdef: :+1: readability

pkg/scheduler/plugin.go
tsenart: I understand now. Will change.

pkg/scheduler/podtask/port_mapping.go
tsenart: OK. I though that didn't matter too much here since there was a public `PodId` before.

pkg/scheduler/podtask/ranges.go
tsenart: Renamed to `Search`.

pkg/scheduler/podtask/ranges_test.go
tsenart: Sorry, misread the code. You're absolutely right. I'll add that test case.

jdef:
We've talked about moving most the generalized port range types and utils up to mesos-go

karlkfi:
@tsenart This code aint here no more... You still want this?

tsenart:
@karlkfi: We can close this I think.

karlkfi:
You want the branch code? If not, delete it.

jdef:
I think this may be worth saving, especially since k8s is going to support port ranges soon. This could could be pretty helpful

jdef:
Yes want to hold onto the branch code

PR #302, opened 2015-05-28
Unify docker build scripts

karlkfi:
- Make build image more like km/complete images
- Use go docker base image
- Change build image name to mesosphere/kubernetes-mesos-build


docker/build/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+        mercurial \
+        git \
+        rsync \
+        patch \
+        && \
+    apt-get clean
+
+COPY ./docker/bin/* /kubernetes-mesos/bin/
+COPY ./docker/build/bin/* /kubernetes-mesos/bin/
+ENV PATH /kubernetes-mesos/bin:$PATH
```
jdef: referencing $PATH works, eh? is /go/bin in the default PATH?

docker/build/bin/km-build
```diff
@@ -8,14 +12,14 @@ set -vx
...
-test -n "$GOPATH" || GOPATH=/pkg
+test -n "$GOPATH" || GOPATH=/go
 
 # gopath is first directory in GOPATH : separated list
 gopath="${GOPATH%%:*}"
 pkg="${gopath}/src/${GOPKG}"
 
 if [ -d $SNAP ]; then
-  test ! -L "${pkg}" || /bin/rm -vf "${pkg}"  # remove any existing link
+  test ! -L "${pkg}" || rm -vf "${pkg}"  # remove any existing link
```
jdef: I think I used /bin/rm out of habit, wary of env aliases. :p

docker/build/bin/km-build
```diff
@@ -8,14 +12,14 @@ set -vx
 GOPKG=github.com/mesosphere/kubernetes-mesos
 
 test ${#} -eq 0 && CMD=( make bootstrap install ) || CMD=( "${@}" )
-test -n "$GOPATH" || GOPATH=/pkg
+test -n "$GOPATH" || GOPATH=/go
```
jdef: pretty sweet if that's the only place you need to change /pkg; what about the project-level Makefile? or hack/patches/apply.sh ?

docker/build/Dockerfile
karlkfi: The golang image handles adding /go/bin to the PATH.

jdef:
LGTM :ship:

jdef:
please squash prior to merge

docker/build/bin/km-build
karlkfi: I've never seen `/pkg` used as a GOPATH before. Usually Go apps don't need to know where the gopath is. So the only things aware are build scripts. I'm pretty sure km-build (renamed from hack/dockerbuild/run.sh) was the only thing using `/pkg`, because that's what the Dockerfile was setting. So it was the only thing I needed to change. 

That said, `km-build` probably shouldn't have a default in it at all. I'll change it to explode if not set. 

FYI: Go generally requires GOPATH to be set (at compilation time, not runtime), but the conventional defaults are `/go` (single-user envs, like docker images) or `~/go` (multi-user envs, like a workstation).

jdef: I think /pkg was something that I cooked up at 3am one morning. /go and ~/go make more sense.

docker/build/bin/km-build
```diff
@@ -8,14 +12,18 @@ set -vx
 GOPKG=github.com/mesosphere/kubernetes-mesos
 
 test ${#} -eq 0 && CMD=( make bootstrap install ) || CMD=( "${@}" )
-test -n "$GOPATH" || GOPATH=/pkg
+
+if [ -z "$GOPATH" ]; then
+  echo "ERROR: GOPATH must be set" >&2
+  exit 1
+fi
```
karlkfi: Changed to error if not set (Dockerfile sets it).

karlkfi:
Squashed and made GOPATH requires to avoid hardcoding `/go` anywhere.

Issue #86, opened 2014-11-28
add optional framework parameter, dockercfg_path


jdef:
when the kubelet-executor spins up it checks for a .dockercfg file. the framework should (optionally) distribute this file if the user provides the path to one at startup time.

jdef:
xref https://github.com/GoogleCloudPlatform/kubernetes/blob/release-0.18/pkg/credentialprovider/config.go#L66

jdef:
fix has been merged, closing out

PR #321, opened 2015-06-03
Add a dockercfg_path parameter to the scheduler

MatMaul:
This is implemented by copying the dockercfg file in the executor work directory.

This handles issue #86.

### DIFF
--- a/pkg/scheduler/service/service.go
+++ b/pkg/scheduler/service/service.go
@@ -87,6 +87,7 @@ type SchedulerServer struct {
 	MesosRole                     string
 	MesosAuthPrincipal            string
 	MesosAuthSecretFile           string
+	DockerCfgPath                 string
 	Checkpoint                    bool
 	FailoverTimeout               float64
 	ExecutorBindall               bool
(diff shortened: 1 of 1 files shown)


pkg/scheduler/service/service.go
```diff
@@ -224,21 +226,30 @@ func (s *SchedulerServer) AddHyperkubeFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.KMPath, "km_path", s.KMPath, "Location of the km executable, may be a URI or an absolute file path.")
 }
 
-// returns (downloadURI, basename(path))
 func (s *SchedulerServer) serveFrameworkArtifact(path string) (string, string) {
+	return s.serveFrameworkArtifactWithFilename(path, "")
+}
+
+// returns (downloadURI, basename(path))
```
sttts: ..., basename(path) or filename if != ""

pkg/scheduler/service/service.go
```diff
@@ -224,21 +226,30 @@ func (s *SchedulerServer) AddHyperkubeFlags(fs *pflag.FlagSet) {
...
+func (s *SchedulerServer) serveFrameworkArtifactWithFilename(path string, filename string) (string, string) {
 	serveFile := func(pattern string, filename string) {
```
sttts: confusing parameter filename now that filename is also a toplevel parameter

sttts:
two small nits, otherwise LGTM

pkg/scheduler/service/service.go
MatMaul: You mean compute the basename ? I was handling the case of an empty basename inside serveFrameworkArtifactWithFilename, but you're right it is better to do that here. In my next pull request I also removed the second return parameter in serveFrameworkArtifactWithFilename now that filename is systematically used (no edge case).

pkg/scheduler/service/service.go
```diff
@@ -224,35 +226,40 @@ func (s *SchedulerServer) AddHyperkubeFlags(fs *pflag.FlagSet) {
...
+		basename = pathSplit[len(pathSplit)-1]
 	} else {
-		base = path
+		basename = path
+	}
+
+	return s.serveFrameworkArtifactWithFilename(path, basename), basename
+}
+
+// returns (downloadURI, basename(path))
```
jdef: docs are now out of date, please update

MatMaul: updated

jdef:
holding off on merge due to upstream efforts

jdef:
needs port to upstream contrib/mesos

MatMaul:
It is already done on my side, I am currently waiting for some answers regarding  the google CLA on my company side. I will send a PR as soon as it is ok.

MatMaul:
upstream PR : kubernetes/kubernetes#12858

PR #357, opened 2015-06-16
Upstreamed all the Go code

jdef:
- removed go code and disabled patches; it is all now upstream
- build kubernetes-mesos from the upstream repo
- remove Godeps and Makefile
- remove travis integration
- remove outdated docs, add links to upstream
- docker/build generates docker image tags using hack/lib/version.sh


jdef:
:ship:

Issue #359, opened 2015-06-17
Docker build container is not working anymore


altvnk:
Due to deletion of all code from kubernetes-mesos repo docker image `mesosphere/kubernetes-mesos-build` is not working for now. Is there any similar way to build framework such simple as it was?

sttts:
This is being worked on for the version in the k8s upstream repo.

altvnk:
Thanks for your answer!
I have one more question - does it possible to build framework via kubernetes build image? [k8s guide](https://github.com/GoogleCloudPlatform/kubernetes/tree/master/build)
I need to export environment variable `KUBERNETES_CONTRIB=mesos` to build framework?

sttts:
This could work. I think nobody tried. `KUBERNETES_CONTRIB=mesos` is correct.

altvnk:
Many thanks! I will try it in short.

altvnk:
Unfortunately that doesn't worked, because somehow `contrib` dir is not copied into docker image and build fails. Trying to investigate that and find a way to build binaries. If it would possible i think then your build image will not be needed at all, all the things will be in one place.

sttts:
That would be great. Looking forward to what you find out. We need this for running e2e tests against a docker compose setup of k8sm.

jdef:
I just pushed a new kubernetes-mesos-build image up to dockerhub:

https://registry.hub.docker.com/u/mesosphere/kubernetes-mesos-build/

altvnk:
Oh that's great! Will test it now.

altvnk:
Worked for me, thank you!

sttts:
```
KUBERNETES_CONTRIB=mesos bash -x build/run.sh hack/build-go.sh
```

builds the k8sm binaries just fine.

altvnk:
I was trying to provide env variable via Dockerfile, that's was my mistake...

PR #361, opened 2015-06-18
[WIP] docker-compose updates

karlkfi:


jdef:
super excited to see this

sttts:
Do we still need the build container? Compare https://github.com/mesosphere/kubernetes-mesos/issues/359#issuecomment-112886908

docker/docker-compose.yml
```diff
@@ -10,93 +10,136 @@ etcd:
...
+  - MESOS_MASTER=mesosmaster1:5050
+  - MESOS_PORT=5051
+  - MESOS_LOG_DIR=/var/log/mesos
+  - MESOS_LOGGING_LEVEL=INFO
+  - MESOS_RESOURCES=cpus:4;mem:1280;disk:25600;ports:[21000-21099]
+  - MESOS_SWITCH_USER=0
+  - MESOS_CONTAINERIZERS=docker,mesos
+  - DOCKER_DAEMON_ARGS="--log-level=error"
+  links:
+  - etcd
```
sttts: the slave doesn't depend on etcd

docker/docker-compose.yml
```diff
@@ -10,93 +10,136 @@ etcd:
...
+  - MESOS_MASTER=mesosmaster1:5050
+  - MESOS_PORT=5052
+  - MESOS_LOG_DIR=/var/log/mesos
+  - MESOS_LOGGING_LEVEL=INFO
+  - MESOS_RESOURCES=cpus:4;mem:1280;disk:25600;ports:[22000-22099]
+  - MESOS_SWITCH_USER=0
+  - MESOS_CONTAINERIZERS=docker,mesos
+  - DOCKER_DAEMON_ARGS="--log-level=error"
+  links:
+  - etcd
```
sttts: the slave doesn't depend on etcd

docker/docker-compose.yml
```diff
@@ -10,93 +10,136 @@ etcd:
...
+    --service-node-port-range=22075-22125
+    --cloud-provider=mesos
+    --cloud-config=/opt/mesos-cloud.conf
     --v=2
   ports: [ "8888:8888" ]
   links:
   - etcd
-  - mesos
+  - mesosmaster1
+  - mesosslave1
```
sttts: the apiserver does not depend on the slaves

docker/docker-compose.yml
```diff
@@ -10,93 +10,136 @@ etcd:
...
     km controller-manager
     --master=http://apiserver:8888
-    --cloud_config=/opt/mesos-cloud.conf
+    --cloud-config=/opt/mesos-cloud.conf
     --v=2
   links:
   - etcd
-  - mesos
+  - mesosmaster1
+  - mesosslave1
```
sttts: the controller does not depend on the slaves

docker/docker-compose.yml
```diff
@@ -10,93 +10,136 @@ etcd:
...
-    await-health-check -t=30 http://apiserver:8888/healthz &&
+    await-health-check http://mesosmaster1:5050/health &&
+    await-health-check -t=60 http://apiserver:8888/healthz &&
     km controller-manager
     --master=http://apiserver:8888
-    --cloud_config=/opt/mesos-cloud.conf
+    --cloud-config=/opt/mesos-cloud.conf
     --v=2
   links:
   - etcd
```
sttts: The controller does not depend on etcd

docker/docker-compose.yml
```diff
@@ -10,93 +10,136 @@ etcd:
...
+    --etcd-servers=http://etcd:4001
+    --mesos-user=root
+    --api-servers=http://apiserver:8888
+    --mesos-master=mesosmaster1:5050
     --v=2
   links:
   - etcd
-  - mesos
+  - mesosmaster1
+  - mesosslave1
```
sttts: the scheduler does not depend on the slaves

docker/test/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+        patch \
+        jq \
+        python \
+        python-pip \
+        && \
+    apt-get clean
+
+RUN curl -sSL https://get.docker.com/ | sh
+
+RUN pip install -U docker-compose
```
sttts: why do we need docker-compose inside the test container?

docker/test/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+        git \
+        rsync \
+        patch \
+        jq \
+        python \
+        python-pip \
+        && \
+    apt-get clean
+
+RUN curl -sSL https://get.docker.com/ | sh
```
sttts: why do we need docker inside the test container?

docker/test/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN pip install -U docker-compose
+
+RUN go get github.com/tools/godep
+
+RUN mkdir -p /go/src/github.com/GoogleCloudPlatform/kubernetes
+WORKDIR /go/src/github.com/GoogleCloudPlatform/kubernetes
+
+COPY ./bin/* /usr/local/bin/
+
+RUN install-etcd.sh
```
sttts: Isn't there an ubuntu package for that?

docker/docker-compose.yml
```diff
@@ -10,93 +10,136 @@ etcd:
...
   links:
   - etcd
   - "ambassador:apiserver"
+mesosslave1:
+  hostname: mesosslave1
+  privileged: true
+  image: mesosphere/kubernetes-mesos-slave:latest
+  ports: [ "5051:5051" ]
+  environment:
+  - MESOS_HOSTNAME=mesosslave1
```
sttts: If you don't set the hostname at all, mesos-slave will use the internal eth0 IP 172.x.y.z for that. Mesos-master then will communicate directly using this IP. This will make the ambassador unnecessary.

docker/docker-compose.yml
```diff
@@ -10,93 +10,136 @@ etcd:
...
+  - MESOS_LOG_DIR=/var/log/mesos
+  - MESOS_LOGGING_LEVEL=INFO
+  - MESOS_RESOURCES=cpus:4;mem:1280;disk:25600;ports:[22000-22099]
+  - MESOS_SWITCH_USER=0
+  - MESOS_CONTAINERIZERS=docker,mesos
+  - DOCKER_DAEMON_ARGS="--log-level=error"
+  links:
+  - etcd
+  - mesosmaster1
+  - "ambassador:apiserver"
```
sttts: Why the ambassador:apiserver in the slave? Using the internal 172.x.y.z IP as advertised apiserver name would allow a direct communication between the two.

sttts:
Created PR #362 to address the dependency issues.

sttts:
Create PR #363 for a working Docker-in-Docker compose yml.

docker/docker-compose.yml
karlkfi: By "at all" do you mean removing line 30 and 35 or just 35?

docker/docker-compose.yml
karlkfi: None of the kublet or executor components need it?

docker/docker-compose.yml
karlkfi: Isn't the apiserver advertising the "apiserver" hostname (rather than the IP)?

docker/docker-compose.yml
karlkfi: In this case, it's using the etcd hostname for the first await-health-check. I could remove that check and just await apiserver, but i'd have to set the timeout to like 2 minutes.

docker/test/Dockerfile
karlkfi: It's outdated, needed a newer one.

docker/test/Dockerfile
karlkfi: The test container is being used to `docker-compose up` on CI so as not to have to install docker-compose, which requires python and pip.

docker/test/Dockerfile
sttts: Ic, so the e2e tests speak with the etcd, right?

docker/test/Dockerfile
sttts: Got it. Makes sense.

sttts:
I suggest we keep using the ambassador and drop #362. 

The best solution would be something like https://github.com/tonistiigi/dnsdock. But there is no elegant way to start that inside docker-compose and pass the dns to all other containers.

docker/test/Dockerfile
karlkfi: The integration tests spin up a local etcd.

karlkfi:
The build script is slightly modified, which is why the cluster/mesos/docker/readme says to build it. It's likely better to build it locally as "latest" for now until we get automated publishing.

karlkfi:
This is obsolete, now that all the images have been migrated to kubernetes: https://github.com/GoogleCloudPlatform/kubernetes/pull/10049

docker/README.md
```diff
@@ -66,7 +66,17 @@ docker kill kubernetes-mesos
...
+It will launch several docker containers linked together with hostnames and port forwarding:
+
+- [docker-grand-ambassador](https://github.com/cpuguy83/docker-grand-ambassador) - Proxy to allow circular hostname dependencies in docker
+- [etcd](https://quay.io/repository/coreos/etcd) - Key/Value store used by Mesos
+- Mesos Master - Single (non-HA) Mesos API server
+- Mesos Slave (x2) - Mesos agents that run Kubernetes Kublets
+- Kubernetes API Server
+- Kubernetes Controller Manager
+- Kubernetes Scheduler
+
```
karlkfi: Details of the containers used should be migrated to the mesos-docker guide.

karlkfi:
This has been superseded by the upstream PR: https://github.com/GoogleCloudPlatform/kubernetes/pull/10049

PR #362, opened 2015-06-18
Remove the need for an ambassador by using IP hostnames in docker-compose setup

sttts:


docker/docker-compose.yml
```diff
@@ -42,32 +31,40 @@ mesosslave1:
   - MESOS_CONTAINERIZERS=docker,mesos
   - DOCKER_DAEMON_ARGS="--log-level=error"
   links:
-  - etcd
   - mesosmaster1
-  - "ambassador:apiserver"
   volumes:
   - /var/run/docker.sock:/var/run/docker.sock
+  - /sys:/sys
+  - /usr/local/bin/docker:/usr/bin/docker
```
karlkfi: Why is the host's docker binary and sys being mounted? Doesn't the container include docker?

docker/docker-compose.yml
```diff
@@ -104,18 +99,14 @@ controller:
   - >
     echo "Hostname: $(hostname -f) ($(hostname -f | xargs resolveip))" &&
     (grep "mesos-master\s*=" /opt/mesos-cloud.conf || echo "  mesos-master = mesosmaster1:5050" >> /opt/mesos-cloud.conf) &&
-    await-health-check http://etcd:4001/health &&
     await-health-check http://mesosmaster1:5050/health &&
```
karlkfi: if you drop the await for etcd, the next await needs an increased timeout

docker/docker-compose.yml
sttts: I have split the dind and non-dind variant. Here the second one is used, i.e. no wrapdocker and hence no mounted cgroups. That's why /sys is mounted.

The problem with installing docker is that docker has changed the API version multiple times. Then your inner docker cannot speak with the outer anymore. This way it can. On the other hand, the outer binary must be at the specific location in the filesystem and be runnable in the container. No free lunch with either way.

PR #390, opened 2015-07-15
Add consecutive 'X's to template of mktemp

nak3:


nak3:
In Linux, TEMPLATE needs to contain at least 3 consecutive 'X's.

> $ man 1 mktemp
> Create  a  temporary  file  or  directory,  safely,  and  print its name.  TEMPLATE must contain at least 3 consecutive 'X's in last component.  If TEMPLATE is not specified, use tmp.XXXXXXXXXX, and
> --tmpdir is implied.  Files are created u+rw, and directories u+rwx, minus umask restrictions.

For example:
~~~
$ mktemp -d -t "k8sm-workspace"
mktemp: too few X's in template ‘k8sm-workspace’
$ mktemp -d -t "k8sm-workspace-XXX"
/tmp/k8sm-workspace-Ox5
~~~

karlkfi:
This was taken into consideration when the upstreaming happened. The docker dir is now obsolete and being deleted.

PR #392, opened 2015-07-16
Add `chcon` to workspace directroy of docker build scripts

nak3:


jdef:
we use the upstream k8s build scripts now for generating binaries. it's not clear to me that this PR is still relevant. closing out.

PR #395, opened 2015-07-23
DCOS: Asset server and reverse proxy

tsenart:
/cc @sttts

### DIFF
--- a/revproxy/.gitignore
+++ b/revproxy/.gitignore
@@ -1,0 +1,1 @@
+revproxy
--- a/revproxy/decorators.go
+++ b/revproxy/decorators.go
@@ -1,0 +1,76 @@
+package main
+
+import (
+	"encoding/json"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// A decorator decorates an http.Handler with a layer of behaviour
+type decorator func(http.Handler) http.Handler
+
+// decorate decorates an http.Handler with all the given decorators, in order.
+func decorate(h http.Handler, ds ...decorator) http.Handler {
+	decorated := h
+	for _, decorate := range ds {
+		decorated = decorate(decorated)
+	}
+	return decorated
+}
+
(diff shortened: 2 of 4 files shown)


jdef:
something to consider: kubectl itself already provides reverse proxy capabilities. I believe that it also handles HTTP to SPDY connection upgrades, which are required for some apiserver endpoints.

have we already exhausted the possibility of using kubectl for proxying?

sttts:
Do you have an idea how to use it for that? Afaik it can only proxy services to make them available on the local host.

jdef:
looks like 127.0.0.1 is hardcoded. however, we could save ourselves some elbow grease and just embed the kubectl.ProxyServer in a struct of our own, and override the Listen() func.

building a ProxyServer looks pretty easy:
https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/kubectl/cmd/proxy.go

and the Listen() func is pretty straightforward to override:
https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/kubectl/proxy_server.go#L181

thoughts?

sttts:
If the code is re-usable, I am all for it. @tsenart has to take a look.

tsenart:
Remember that besides the proxy, we need to serve files too. But yeah, I don't want to write SPDY upgrade code if I don't have to... So, I'm +1 on reusing that code for the proxy function.

tsenart:
I won't be able to work on this until next week, so feel free to take over if it's a priority.

karlkfi:
@tsenart What's the status of this? Should it be upstream in contrib? I know there was work recently on a reverse proxy that was demoed at a recent k8s sync.

sttts:
I would like to approach this topic again in the next sprint.

sttts:
Superseeded by https://github.com/mesosphere/kubernetes-mesos/pull/519. Closing.

PR #397, opened 2015-07-24
Clean up and add kube-ui addon

sttts:


hack/images/dcos/Makefile
```diff
@@ -46,8 +48,11 @@ etcd:
 $(ETCD_BINS): etcd
 	test -f "$$(basename $@)" || cp -v $@ .
 
-$(KUBE_DNS_TEMPLATES): $(KUBE_DNS_TEMPLATES_SOURCES)
-	@cp -v ../dns/$@ .
+$(KUBE_DNS_TEMPLATES):
+	curl -f https://raw.githubusercontent.com/GoogleCloudPlatform/kubernetes/$(REF)/cluster/addons/dns/$@ > $@
+
+$(KUBE_UI_TEMPLATES):
+	curl -f https://raw.githubusercontent.com/GoogleCloudPlatform/kubernetes/$(REF)/cluster/addons/kube-ui/$@ > $@
```
tsenart: These two `curls` look very similar. Would it be possible to refactor?

hack/images/dcos/bootstrap.sh
```diff
@@ -22,7 +22,7 @@ echo
 # -C for failing when files are clobbered
 set -ue
 
-if [ "${DEBUG}" == "true" ]; then
+if [ "${DEBUG:-false}" == "true" ]; then
```
tsenart: Is this default added so that bash doesn't complain about an unset variable?

hack/images/dcos/bootstrap.sh
```diff
@@ -71,6 +71,12 @@ controller_manager_host=${CONTROLLER_MANAGER_HOST:-${default_dns_name}}
 controller_manager_port=${CONTROLLER_MANAGER_PORT:-10252}
 echo "* controller manager: $controller_manager_host:$controller_manager_port"
 
+enable_dns=${ENABLE_DNS:-true}
```
tsenart: Why create a similarly named variable? If it is unset, we can set a default to it: `: ${ENABLE_DNS="true"}`

hack/images/dcos/bootstrap.sh
sttts: Right.

tsenart: I'd rather set a default once instead of having to do this everywhere. `: ${DEBUG=false}`

sttts: done.

hack/images/dcos/bootstrap.sh
sttts: done.

hack/images/dcos/bootstrap.sh
```diff
@@ -220,70 +218,88 @@ exec 2>&1
...
-  prepare_service_depends apiserver ${kube_master}/healthz ok kube_dns
+  apiserver_depends="${apiserver_depends} kube_dns"
 }
 
+# launch kube-dns if enabled
 kube_cluster_dns=""
 kube_cluster_domain=""
-# launch kube-dns if enabled
-if test -n "$ENABLE_DNS"; then
+if [ "${ENABLE_DNS:-true}" == true ]; then
```
tsenart: Same as before for this case.

sttts: What do you mean? I kept the style to pass the default in place. One could that once for all variables at the top. But that's another refactoring.

hack/images/dcos/bootstrap.sh
```diff
@@ -220,70 +218,88 @@ exec 2>&1
...
+  /opt/kubectl get service --namespace=kube-system kube-ui >/dev/null && \
+  touch kill && exit 0
 
-failover_timeout="${K8SM_FAILOVER_TIMEOUT:-}"
-if test -n "$failover_timeout"; then
-  failover_timeout="--failover-timeout=$failover_timeout"
+/opt/kubectl create -f /opt/kube-ui-rc.yaml
+/opt/kubectl create -f /opt/kube-ui-svc.yaml
+EOF
+  sed -i -e '$i test -f kill && exec s6-svc -d $(pwd) || exec \\' ${service_dir}/kube_ui/finish
```
tsenart: My eyes started to glaze over this :)

hack/images/dcos/bootstrap.sh
```diff
@@ -220,70 +218,88 @@ exec 2>&1
...
+  apiserver_depends="${apiserver_depends} kube_ui"
+}
+
+if [ "${ENABLE_UI:-true}" == true ]; then
```
tsenart: Same as before.

hack/images/dcos/bootstrap.sh
sttts: Was copied and pasted from the kube-dns. Better delete it :)

sttts: In fact there might be a sense in this line. This whole service generation must be cleaned up. It's far too complicated. But I wouldn't do that in this PR.

tsenart:
LGTM

PR #419, opened 2015-07-29
Make DISABLE_ETCD_SERVER a boolean

tsenart:
Ref: https://github.com/mesosphere/universe/pull/184

/cc @jdef @sttts


jdef:
can we also change:
```
kube_cluster_domain=${DNS_DOMAIN:-kubernetes.local}
```
to default to cluster.local, which is the default value in the universe repo (or will be?)

sttts:
done.

PR #519, opened 2015-09-30
Start nginx on APISERVER_PROXY_PORT in front of apiserver

sttts:
- kubectl binaries are built for Linux, Mac and Windows from the Makefile
- kubectl binaries are copied to /opt/static in the container
- the `nginx` process
  - serves the binaries under the `/static` path
  - redirects /ui to the kube-ui URL
  - proxy everything else to the apiserver.

- [x] provide meta information about kubectl binaries which include a sha1/md5 of the binaries
- [x] adapt the k8s subcommand to download the right binaries and to poll for new versions
- [x] clean-up


images/dcos/Makefile
```diff
@@ -32,16 +32,18 @@ DOCKER_IMAGE = $(DOCKER_REPO):$(VERSION)$(POSTFIX)
...
 endif
+OS_EXT = $(if $(filter $(1),windows),.exe,)
 K8S_BINARIES_SOURCE_DIR = $(K8S_ROOT)/_output/dockerized/bin/$(1)/amd64
-K8S_BINARIES_SOURCE = $(addprefix $(call K8S_BINARIES_SOURCE_DIR,linux)/,$(K8S_BINARIES))
-K8S_BINARIES_DEST = $(addprefix $(BUILD_DIR)/,$(K8S_BINARIES))
-K8S_KUBECTL_SOURCE = $(addsuffix /kubectl,$(call K8S_BINARIES_SOURCE_DIR,linux) $(call K8S_BINARIES_SOURCE_DIR,darwin))
+K8S_KUBECTL_OSES = linux windows darwin
+K8S_KUBECTL_SOURCE = $(foreach OS,$(K8S_KUBECTL_OSES),$(call K8S_BINARIES_SOURCE_DIR,$(OS))/kubectl$(call OS_EXT,$(OS)))
+K8S_KUBECTL_DEST = $(foreach OS,$(K8S_KUBECTL_OSES),$(BUILD_DIR)/$(OS)/kubectl$(call OS_EXT,$(OS)))
+K8S_BINARIES_SOURCE = $(K8S_KUBECTL_SOURCE) $(addsuffic /km,$(call K8S_BINARIES_SOURCE_DIR,linux))
```
jdef: s/suffic/suffix/

sttts: fixed

images/dcos/bootstrap.sh
```diff
@@ -51,9 +51,11 @@ default_dns_name=${DEFAULT_DNS_NAME:-k8sm.marathon.mesos}
 echo "* DNS name: $default_dns_name"
 
 apiserver_host=${APISERVER_HOST:-${default_dns_name}}
-apiserver_port=${APISERVER_PORT:-8888}
+apiserver_port=18888
```
jdef: TODO: this should be overridable by k8s/dcos packaging

sttts: yeah, this is still quick'n'dirty. still testing this on dcos.

images/dcos/bootstrap.sh
```diff
@@ -194,11 +199,27 @@ ${apply_uids}
...
+#
+prepare_service ${monitor_dir} ${service_dir} controller-manager ${CONTROLLER_MANAGER_RESPAWN_DELAY:-3} <<EOF
+#!/usr/bin/execlineb
+fdmove -c 2 1
+${apply_uids}
+/opt/kubectl proxy
+  --server=${kube_master}
+  --www=/opt/static
+  --bind-address=${host_ip}
+  --port=${proxy_kube_master}
```
jdef: this looks strange, passing an ip:port to a --port flag

images/dcos/bootstrap.sh
```diff
@@ -194,11 +199,27 @@ ${apply_uids}
   --address=${host_ip}
   --cloud-config=${cloud_config}
   --cloud-provider=mesos
-  --master=http://${host_ip}:${apiserver_port}
+  --master=${host_kube_master}
```
jdef: can we just have this use the same value as the scheduler and get rid of this envvar?

images/dcos/bootstrap.sh
```diff
@@ -121,7 +118,15 @@ cat <<EOF >${cloud_config}
 EOF
 
 # address of the apiserver
-kube_master="http://${apiserver_host}:${apiserver_port}"
+kube_master="http://${apiserver_host}:${proxy_apiserver_port}"
```
jdef: can we call this something like `proxied_kube_master`?

images/dcos/bootstrap.sh
```diff
@@ -121,7 +118,15 @@ cat <<EOF >${cloud_config}
 EOF
 
 # address of the apiserver
-kube_master="http://${apiserver_host}:${apiserver_port}"
+kube_master="http://${host_ip}:${apiserver_port}"
+proxy_kube_master="http://${apiserver_host}:${proxy_apiserver_port}"
```
jdef: can we call this something like `public_kube_master`?

images/dcos/bootstrap.sh
```diff
@@ -121,7 +118,15 @@ cat <<EOF >${cloud_config}
 EOF
 
 # address of the apiserver
-kube_master="http://${apiserver_host}:${apiserver_port}"
+kube_master="http://${host_ip}:${apiserver_port}"
```
jdef: call this `cluster_kube_master`?

images/dcos/bootstrap.sh
```diff
@@ -121,7 +118,15 @@ cat <<EOF >${cloud_config}
...
+proxy_kube_master="http://${apiserver_host}:${proxy_apiserver_port}"
+host_kube_master="http://${host_ip}:${apiserver_port}"
```
jdef: call this `private_kube_master`?

sttts:
```bash
$ curl -i 'http://stefansch-elasticl-1ibl7vd1x736k-786726734.eu-central-1.elb.amazonaws.com/service/kubernetes/static/'
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Date: Wed, 30 Sep 2015 14:38:34 GMT
Last-Modified: Wed, 30 Sep 2015 14:36:39 GMT
Server: openresty/1.7.10.2
Content-Length: 101
Connection: keep-alive

<pre>
<a href="linux/">linux/</a>
<a href="darwin/">darwin/</a>
<a href="windows">windows</a>
</pre>
```

images/dcos/bootstrap.sh
jdef: i guess we can't rid of this since proxy needs it so labeling it as `private` and adding a comment that ONLY the kubectl proxy should use it is probably the best we can do

sttts:
@jdef: Why do we use host network mode for our uber package? Could we switch?  I don't like all the host ports.

jdef:
plan is to switch once ip-per-container support lands. until then we're
trying to maintain the same port in case the scheduler dies and moves to
another host. once etcd-mesos is in place and apiserver is using that we'll
have a better framework-container failover story .. unless the scheduler
ip:port changes. if the scheduler ip:port changes then the executor URI's
change. and we end up with incompat executors across the cluster. same goes
for the apiserver port that's referenced by the executor.

sttts:
@jdef PTAL. It's not finished yet, still a prototype. Next step: take care of your old review comments.

images/dcos/bootstrap.sh
sttts: fixed

images/dcos/bootstrap.sh
sttts: done

images/dcos/bootstrap.sh
sttts: It's `kube_master_proxy` now.

images/dcos/bootstrap.sh
sttts: Made everything explicit now: `kube_master` is the kube-master. `kube_master_proxy` is the kube-master proxy. Should be clearer.

sttts:
@jdef PTAL, cleaned up, addressed your old comments. Should be ready now.

images/dcos/root/etc/nginx/conf.d/default.conf.template
```diff
@@ -0,0 +1,23 @@
...
+    listen <PORT>;
+    server_name "";
+
+    root /usr/html;
+    index index.html;
+
+    location = / {
+        return 301 $scheme://$http_host/api/v1/proxy/namespaces/kube-system/services/kube-ui/;
+    }
+
```
jdef: don't we need a separate rule for `/api`? I don't have nginx experience and I'm confused as to how this config actually works. Does this say that: (a) when the full URL path is `/` then send a redirect; (b) proxy everything else under `/` to APISERVER, other than for the rules that follow?

If so, some comments would be nice

images/dcos/root/etc/nginx/nginx.conf
```diff
@@ -0,0 +1,30 @@
+user nginx;
```
jdef: do we need to create this user, or does the apk do that for us?

images/dcos/root/usr/html/index.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
```
jdef: why do we need this file? if we really need something here can we rip out this boiler-plate and replace it with something minimal, e.g. `<html><body>k8s/mesos master proxy</body></html>`

images/dcos/Dockerfile
```diff
@@ -1,25 +1,32 @@
-FROM busybox:ubuntu-14.04
+FROM alpine:3.2
```
jdef: we've experienced lots of inconsistency w/ respect to dns resolution across different base containers. curious how this choice will impact us.

/cc @spacejam @karlkfi

images/dcos/root/etc/nginx/conf.d/default.conf.template
sttts: It selects the longest matching rules and then follows the order of definition.

images/dcos/root/etc/nginx/nginx.conf
sttts: apk does that.

images/dcos/root/usr/html/index.html
sttts: We don't need it. Will remove it.

images/dcos/Dockerfile
sttts: I noticed differences already, I think in connection with multiple nameservers, one returning an error. But I think this was due to the nginx compiled against ulibc, not glibc. The image has both, the go binaries link against glibc.

sttts: An alternative would be busybox with the nginx binary from the Ubuntu packages. No idea how man zillions of dependencies this means.

karlkfi: I prefer alpine just because it has a package manager, but I haven't tested DNS resolution.

images/dcos/root/opt/functions.sh
```diff
@@ -3,7 +3,7 @@
 test -n "$sandbox" || die failed to identify mesos sandbox
 
 lookup_ip() {
-  local leader=$(nslookup "$1" | sed -e '/^Server:/,/^Address .*$/{ d }' -e '/^$/d'|grep -e '^Address '|cut -f3 -d' '|head -1)
+  local leader=$(nslookup "$1" | grep -v 127.0.0.1 | sed -e '/^Server:/,/^Address .*$/{ d }' -e '/^$/d'|grep -e '^Address '|cut -f3 -d' '|head -1)
```
karlkfi: is it possible that this returns more than one line and needs to have a `tail -1` in there somewhere?

karlkfi:
What was `webhook.sh` and why was it deleted?

Otherwise no real complaints. LGTM.

jdef:
webhook.sh was an experiment that I was cooking up but never ended up
using. thinking was that the s6 monitor scripts could fire events, via
HTTP, that some (as yet unknown) dcos event service would be listening to
and operators would be able to tell if services were flapping. dcos still
has nothing like this and we're eventually going to split up the uber
container anyway - so it probably makes sense to kill the webhooks file
since it adds no value and it not used at all.

images/dcos/root/opt/functions.sh
sttts: Maybe. This line is a bad hack. We need something more stable. It broke. So I quick fixed it excluding the 127.0.0.1.

PR #556, opened 2015-10-13
use the built-in etcd proxy to use etcd-mesos if configured

spacejam:
@jdef

### DIFF
--- a/images/dcos/bootstrap.sh
+++ b/images/dcos/bootstrap.sh
@@ -145,7 +145,7 @@ exec /opt/etcd \\
   -name ${etcd_server_name}
 EOF
 
-  local deps="controller-manager scheduler"
+  local deps="scheduler"
   if [ -n "${apiserver_depends}" ]; then
     deps="${deps} apiserver-depends"
   else
@@ -169,7 +169,7 @@ exec /opt/etcd \\
   -discovery-srv=${ETCD_MESOS_FRAMEWORK_NAME}.mesos \\
 EOF
 
-  local deps="controller-manager scheduler"
+  local deps="scheduler"
   if [ -n "${apiserver_depends}" ]; then
     deps="${deps} apiserver-depends"
   else
--- a/images/dcos/bootstrap.sh
+++ b/images/dcos/bootstrap.sh
@@ -129,17 +129,17 @@ EOF
 prepare_etcd_service() {
   mkdir -p ${etcd_server_data}
   prepare_service ${monitor_dir} ${service_dir} etcd-server ${ETCD_SERVER_RESPAWN_DELAY:-1} << EOF
-#!/bin/sh
+#!/usr/bin/execlineb
 #TODO(jdef) don't run this as root
 #TODO(jdef) would be super-cool to have socket-activation here so that clients can connect before etcd is really ready
-exec 2>&1
-exec /opt/etcd \\
-  -advertise-client-urls ${etcd_advertise_client_urls} \\
-  -data-dir ${etcd_server_data} \\
-  -initial-advertise-peer-urls ${etcd_initial_advertise_peer_urls} \\
-  -initial-cluster ${etcd_server_name}=${etcd_initial_advertise_peer_urls} \\
-  -listen-client-urls ${etcd_listen_client_urls} \\
-  -listen-peer-urls ${etcd_listen_peer_urls} \\
+fdmove -c 2 1
+/opt/etcd
+  -advertise-client-urls ${etcd_advertise_client_urls}
+  -data-dir ${etcd_server_data}
+  -initial-advertise-peer-urls ${etcd_initial_advertise_peer_urls}
+  -initial-cluster ${etcd_server_name}=${etcd_initial_advertise_peer_urls}
+  -listen-client-urls ${etcd_listen_client_urls}
+  -listen-peer-urls ${etcd_listen_peer_urls}
   -name ${etcd_server_name}
 EOF
 
@@ -155,16 +155,16 @@ EOF
 prepare_etcd_proxy() {
   mkdir -p ${etcd_server_data}
   prepare_service ${monitor_dir} ${service_dir} etcd-server ${ETCD_SERVER_RESPAWN_DELAY:-1} << EOF
-#!/bin/sh
+#!/usr/bin/execlineb
 #TODO(jdef) don't run this as root
 #TODO(jdef) would be super-cool to have socket-activation here so that clients can connect before etcd is really ready
-exec 2>&1
-exec /opt/etcd \\
-  -proxy=on \\
-  -advertise-client-urls ${etcd_advertise_client_urls} \\
-  -data-dir ${etcd_server_data} \\
-  -listen-client-urls ${etcd_listen_client_urls} \\
-  -discovery-srv=${ETCD_MESOS_FRAMEWORK_NAME}.mesos \\
+fdmove -c 2 1
+/opt/etcd
+  -proxy=on
+  -advertise-client-urls ${etcd_advertise_client_urls}
+  -data-dir ${etcd_server_data}
+  -listen-client-urls ${etcd_listen_client_urls}
+  -discovery-srv=${ETCD_MESOS_FRAMEWORK_NAME}.mesos
 EOF
 
   local deps="scheduler"
--- a/images/dcos/bootstrap.sh
+++ b/images/dcos/bootstrap.sh
@@ -79,7 +79,7 @@ echo "* controller manager: $controller_manager_host:$controller_manager_port"
 etcd_server_port=${ETCD_SERVER_PORT:-4001}
 etcd_server_peer_port=${ETCD_SERVER_PEER_PORT:-4002}
 
-ETCD_MESOS=${ETCD_MESOS:-disabled}
+ETCD_MESOS_FRAMEWORK_NAME=${ETCD_MESOS_FRAMEWORK_NAME:-disabled}
 etcd_advertise_server_host=${ETCD_ADVERTISE_SERVER_HOST:-127.0.0.1}
 etcd_server_host=${ETCD_SERVER_HOST:-127.0.0.1}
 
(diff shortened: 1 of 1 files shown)


images/dcos/bootstrap.sh
```diff
@@ -163,6 +154,32 @@ EOF
...
 }
 
+prepare_etcd_proxy() {
+  prepare_service ${monitor_dir} ${service_dir} etcd-server ${ETCD_SERVER_RESPAWN_DELAY:-1} << EOF
+#!/bin/sh
+#TODO(jdef) don't run this as root
+#TODO(jdef) would be super-cool to have socket-activation here so that clients can connect before etcd is really ready
+exec 2>&1
+mkdir -p ${etcd_server_data}
+PATH="/opt:${PATH}"
```
sttts: Is this path really necessary?

images/dcos/bootstrap.sh
```diff
@@ -71,11 +71,6 @@ controller_manager_host=${CONTROLLER_MANAGER_HOST:-${default_dns_name}}
 controller_manager_port=${CONTROLLER_MANAGER_PORT:-10252}
 echo "* controller manager: $controller_manager_host:$controller_manager_port"
 
-# assume that the leading mesos master is always running a marathon
-# service proxy, perhaps using haproxy.
-service_proxy=${SERVICE_PROXY:-leader.mesos}
```
sttts: Why is this removed?

images/dcos/bootstrap.sh
```diff
@@ -84,23 +79,19 @@ echo "* service proxy: $service_proxy"
 etcd_server_port=${ETCD_SERVER_PORT:-4001}
 etcd_server_peer_port=${ETCD_SERVER_PEER_PORT:-4002}
 
-ENABLE_ETCD_SERVER=${ENABLE_ETCD_SERVER:-false}
-if [ "${ENABLE_ETCD_SERVER}" != true ]; then
-  etcd_server_list=${ETCD_SERVER_LIST:-http://${service_proxy}:${etcd_server_port}}
-else
-  etcd_advertise_server_host=${ETCD_ADVERTISE_SERVER_HOST:-127.0.0.1}
-  etcd_server_host=${ETCD_SERVER_HOST:-127.0.0.1}
+ETCD_MESOS=${ETCD_MESOS:-disabled}
```
sttts: This variable name is not really intuitive. How about EXTERNAL_ETCD or something in this direction?

images/dcos/bootstrap.sh
```diff
@@ -163,6 +154,32 @@ EOF
   prepare_service_depends etcd-server ${etcd_server_list}/v2/stats/store getsSuccess ${deps}
 }
 
+prepare_etcd_proxy() {
+  prepare_service ${monitor_dir} ${service_dir} etcd-server ${ETCD_SERVER_RESPAWN_DELAY:-1} << EOF
+#!/bin/sh
+#TODO(jdef) don't run this as root
+#TODO(jdef) would be super-cool to have socket-activation here so that clients can connect before etcd is really ready
+exec 2>&1
```
sttts: All other service use fdmove instead.

images/dcos/bootstrap.sh
```diff
@@ -163,6 +154,32 @@ EOF
...
+mkdir -p ${etcd_server_data}
```
sttts: We don't need this mkdir inside of the service script. Move it outside to simplify the service script.

images/dcos/bootstrap.sh
```diff
@@ -163,6 +154,32 @@ EOF
...
+export PATH
+exec /opt/etcd \\
+  -proxy=on \\
+  -advertise-client-urls ${etcd_advertise_client_urls} \\
+  -data-dir ${etcd_server_data} \\
+  -listen-client-urls ${etcd_listen_client_urls} \\
+  -discovery-srv=${ETCD_MESOS}.mesos \\
+EOF
+
+  local deps="controller-manager scheduler"
```
sttts: Does controller-amager depend on etcd?

images/dcos/bootstrap.sh
spacejam: I could not find a reference to it, and @jdef thought it may be safe to remove.  Do you know otherwise?

sttts: You are right. I guess it comes from the marathon service haproxy in an ancient DCOS version.

images/dcos/bootstrap.sh
jdef: changes in this boostrap file are etcd-mesos specific. the name ETCD_MESOS aligns closely with the dcos config option name that we're adding in multiverse. the goal is to keep the multiverse config options as simple as possible, not to make this bootstrap image as flexible as possible. `disabled` will run an internal etcd server in the uber container. any other value is the name of the etcd-mesos framework that we should attach to.

images/dcos/bootstrap.sh
jdef: fdmove is part of execlineb. this service script uses `/bin/sh`.

images/dcos/bootstrap.sh
jdef: disagree. goal should be to keep things the service needs to start in the service script.

images/dcos/bootstrap.sh
jdef: probably not -- copy/paste from prepare_etcd_server

images/dcos/bootstrap.sh
jdef: these deps were copied from prepare_etcd_server. at this point the controller manager should not have a direct dependency to etcd so we should be safe to remove it from both this script and prepare_etcd_server.

images/dcos/bootstrap.sh
sttts: I understand what it does. But the name ETCD_MESOS does not suggest what it does. What is etcd-mesos in the context of a mesos framework?

jdef: if `$ETCD_MESOS != "disabled"` then you'll want to set `etcd_server_list` to `http://127.0.0.1:$someport` because it's used (a) as an endpoint for etcd readiness, and (b) by other apps that want to connect to etcd (apiserver, scheduler)

jdef: it's a dependency

jdef: we can add some short description here outlining the dependency.
unfortunately dependencies in DCOS are currently less than ideal.

images/dcos/bootstrap.sh
sttts: that's the point. Why are all other scripts using execlineb and this is not?

images/dcos/bootstrap.sh
sttts: I think you miss my point. I don't argue against the variable, just against the bad name, which is neither a verb nor an adjective, but only two context-less words.

images/dcos/bootstrap.sh
jdef: using env variables in execlineb sucks. if we get rid of PATH then we could
migrate to execlineb

images/dcos/bootstrap.sh
jdef: sttts and I settled on ETCD_MESOS_FRAMEWORK_NAME

spacejam: sounds good to me :+1:

spacejam: `etcd_server_list` is always set, whether the proxy is listening there or the embedded server when ETCD_MESOS_FRAMEWORK_NAME is disabled.  If we want to support external clusters in the future we will want to conditionally set this, but that's out of scope for this change.

images/dcos/bootstrap.sh
sttts: @spacejam ^^

spacejam: This detail is uninteresting to me.

sttts: Which means? Nobody knows whether this PATH is needed? Then drop it.

images/dcos/bootstrap.sh
sttts: It's preparation of the environment. It does not change during the runtime. If it wasn't configurable it would even belong into the Dockerfile. But creating directories for persistent data does not belong into a service script for sure.

spacejam:
@sttts I've nuked the PATH stuff and moved the mkdir outside of the service definition, and switched over to execlineb for consistency.

images/dcos/bootstrap.sh
```diff
@@ -136,25 +127,47 @@ prepare_service_script ${service_dir} .s6-svscan finish <<EOF
...
-export PATH
-exec /opt/etcd \\
-  -advertise-client-urls ${etcd_advertise_client_urls} \\
-  -data-dir ${etcd_server_data} \\
-  -initial-advertise-peer-urls ${etcd_initial_advertise_peer_urls} \\
-  -initial-cluster ${etcd_server_name}=${etcd_initial_advertise_peer_urls} \\
-  -listen-client-urls ${etcd_listen_client_urls} \\
-  -listen-peer-urls ${etcd_listen_peer_urls} \\
+fdmove -c 2 1
+/opt/etcd
```
sttts: :+1:

images/dcos/bootstrap.sh
```diff
@@ -71,11 +71,6 @@ controller_manager_host=${CONTROLLER_MANAGER_HOST:-${default_dns_name}}
 controller_manager_port=${CONTROLLER_MANAGER_PORT:-10252}
 echo "* controller manager: $controller_manager_host:$controller_manager_port"
 
-# assume that the leading mesos master is always running a marathon
-# service proxy, perhaps using haproxy.
-service_proxy=${SERVICE_PROXY:-leader.mesos}
-echo "* service proxy: $service_proxy"
```
sttts: Can we add similar echo's again printing out which services are created?

sttts:
Only one comment left about echos. Otherwise, lgtm.

spacejam:
ping!

images/dcos/bootstrap.sh
spacejam: functions.sh:71
```
prepare_service() {         
...                                                                                                                                                                           
  local name="$3"                                                                                                                                                                                                                                                                                                                                     ...                         
  echo "* prepare service ${name}"  
...
}
```

Is this not sufficient?

sttts: You are right, didn't remember there was a generic log output.

jdef:
:ship:

PR #670, opened 2015-12-05
Add APISERVER_RUNTIME_CONFIG variable

sttts:
As a cluster admin I want to configure the availability of alpha and beta APIs or disable old APIs at some point.

### DIFF
--- a/images/dcos/bootstrap.sh
+++ b/images/dcos/bootstrap.sh
@@ -197,6 +197,7 @@ ${apply_uids}
   --secure-port=${apiserver_secure_port}
   --service-cluster-ip-range=${SERVICE_CLUSTER_IP_RANGE:-10.10.10.0/24}
   --v=${APISERVER_GLOG_v:-${logv}}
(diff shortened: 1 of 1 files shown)


images/dcos/bootstrap.sh
```diff
@@ -197,6 +197,7 @@ ${apply_uids}
   --secure-port=${apiserver_secure_port}
   --service-cluster-ip-range=${SERVICE_CLUSTER_IP_RANGE:-10.10.10.0/24}
   --v=${APISERVER_GLOG_v:-${logv}}
+  --runtime-config=${APISERVER_RUNTIME_CONFIG:-}
```
jdef: other variables that we expect external things to populate have defaults. why not this? if there is no reasonable default, then why not follow https://github.com/mesosphere/kubernetes-mesos/pull/670/files#diff-2209c7c80934f1bac7d121b3a61d7d51L335 ?

jdef:
motivation: to be able to experiment with new/up-and-coming apis

Issue #535, opened 2015-10-05
Use https from kubectl through the admin router to the apiserver


sttts:
As a user I want to
- have kubectl exec & log to work
- have a secure connection between my local machine and the apiserver.

TODO:
- [x] implement TLS support for the apiserver
- [x] look into the admin router implementation
- [x] prototype a config change to allow SPDY

sttts:
cc @karlkfi

sttts:
If I understand the SPDY support correctly, SPDY depends on HTTPS. To get HTTPS in DCOS we need certificate generation and probably service accounts.

sttts:
Not sure I understood the CM.3 demo correctly: does it bring SSL admin router support?

sttts:
This is blocked until we deploy TLS, service accounts etc.

jdef:
blocked by #555

PR #714, opened 2015-12-22
Implement TLS support

sttts:
- [x] use TLS from kubectl to admin router, from admin router to uber container nginx and from that to the apiserver, each with self-signed snakeoil certificates
- [x] add SPDY upgrade support
- [x] add sandbox overlay with nsenter and socat binaries
- [x] fix duplicate nobody user bug in uber container
- [x] add HTTPS support for the admin router: https://github.com/mesosphere/adminrouter/pull/17
- [x] add HTTPS support to dcos-kubectl: https://github.com/mesosphere/dcos-kubectl/releases/tag/v0.2.1
- [x] test & analyse long running connection termination (e.g. in kubectl-exec): pinging the connection with some output help. This is known: https://github.com/kubernetes/kubernetes/pull/18552, https://github.com/kubernetes/kubernetes/issues/12179

Fixes https://github.com/mesosphere/kubernetes-mesos/issues/535
Follow-up in https://github.com/mesosphere/kubernetes-mesos-private/issues/9, commit also in this branch.


images/dcos/bootstrap.sh
```diff
@@ -220,7 +220,7 @@ EOF
 #
 # nginx, proxying the apiserver and serving kubectl binaries
 #
-sed "s,<PORT>,${apiserver_proxy_port},;s,<APISERVER>,http://${host_ip}:${apiserver_port}," /etc/nginx/conf.d/default.conf.template > /etc/nginx/conf.d/default.conf
+sed "s,<PORT>,${apiserver_proxy_port},;s,<APISERVER>,https://${host_ip}:${apiserver_secure_port}," /etc/nginx/conf.d/default.conf.template > /etc/nginx/conf.d/default.conf
```
jdef: in a comment on the related SSL/TLS "investigation" ticket you mentioned that https wasn't required for SPDY upgrade. so why the additional overhead of using https here?

images/dcos/root/etc/nginx/conf.d/default.conf.template
```diff
@@ -1,5 +1,13 @@
 server {
-    listen <PORT>;
+    listen <PORT> ssl spdy;
+    ssl_certificate snakeoil.crt;
```
jdef: would be nice to allow the user to override via config-prop/envvar's like `kubernetes.certificate.public` and `kubernetes.certificate.private`, so as to avoid the checked-in (thereby insecure) defaults. it wouldn't hurt the size of our bootstrap too much to incorporate a few envvar checks. i think the marathon-lb project uses envvars for this. https://github.com/mesosphere/marathon-lb#providing-ssl-certificates

jdef:
excited to see progress on this front. some minor comments.

images/dcos/bootstrap.sh
sttts: AFAIK Nginx does not support SPDY without TLS. But it would be worth further investigation.

In the other comment I meant that k8s is able to upgrade a HTTP connection to SPDY.

images/dcos/root/etc/nginx/conf.d/default.conf.template
sttts: Yeah, that's the plan.

s-urbaniak:
LGTM

images/dcos/root/etc/nginx/conf.d/default.conf.template
```diff
@@ -11,6 +19,25 @@ server {
...
+        # Authenticate
+        proxy_set_header Authorization "Bearer <TOKEN>";
+
+        # Don't try to be intelligent, only proxy
+        proxy_buffering off;
+        proxy_redirect off;
+        proxy_next_upstream off;
+
+        # Set long timeout
+        proxy_connect_timeout 159s;
```
jdef: where do these timeouts come from (how were they determined)?

jdef:
minor comment re: timeouts, otherwise lgtm

images/dcos/root/etc/nginx/conf.d/default.conf.template
sttts: There is no heartbeating in kubectl-exec and other commands. So the connection is broken eventually without traffic. There are open GH issues upstream (kubernetes/kubernetes#18552, kubernetes/kubernetes#12179). I think I took the same values as the nginx in the GCE salt scripts. But these do not fix the actual problem.

images/dcos/Dockerfile
```diff
@@ -9,14 +9,16 @@ RUN apk --update add nginx curl ca-certificates && \
...
 RUN curl -L "https://github.com/just-containers/s6-overlay/releases/download/$S6_VERSION/s6-overlay-amd64.tar.gz" | tar xzvf - -C /
 
-RUN echo 'nobody:x:99:99:Nobody:/:/sbin/false' >>/etc/passwd
-RUN echo 'nobody::99:' >>/etc/group
-
 # fix nginx tmp directory permissions to be writable by nginx user
 RUN mkdir -p /var/lib/nginx/tmp && chown nginx /var/lib/nginx /var/lib/nginx/tmp && chmod 0700 /var/lib/nginx /var/lib/nginx/tmp
 
+# fix /var/run/kubernetes permissions such that the apisever can create
+# self-signed certificates
```
jdef: does "self-signed certificates" really apply any more now that we have static certs? @sttts

Issue #790, opened 2016-02-18
etcd_proxy consuming unallocated port in k8s/dcos


jdef:
user reported port conflict on a slave node because etcd_proxy was consuming a port that hadn't been properly allocated:
```
E0217 13:27:18.112458 6688 manager.go:1864] Failed to create pod infra container: API error (500): Cannot start container 4c74ded9de8418510f0c7966faef1a5adc834029b3e35829744bdcf6ee783a47: failed to create endpoint k8s_POD.e67a4f09_pod-38_default_50f9670b-d5a3-11e5-8fb3-005056ad6f59_e60cef0d on network bridge: Bind for 0.0.0.0:1052 failed: port is already allocated 
```

jdef:
running k8s-mesos with etcd-mesos I see the following:
```
core@ip-10-0-2-150 ~ $ sudo netstat -natlp |grep -e etcd|grep -e LISTEN
tcp        0      0 127.0.0.1:4001          0.0.0.0:*               LISTEN      1933/etcd
tcp        0      0 10.0.2.150:1025         0.0.0.0:*               LISTEN      1264/./etcd
tcp        0      0 10.0.2.150:1026         0.0.0.0:*               LISTEN      1264/./etcd
tcp        0      0 10.0.2.150:38594        0.0.0.0:*               LISTEN      1250/./etcd-mesos-e
tcp6       0      0 :::1027                 :::*                    LISTEN      1250/./etcd-mesos-e
```

* etcd-mesos isn't properly allocating a driver port from the resource pool
* the etcd proxy running in the k8sm uber container is using local port 4001, not properly allocating that from the resource pool; since the container uses HOST networking mode this could create conflicts

jdef:
going to change the k8s/dcos scheduler docker container to run in BRIDGE mode, requires some changes in the scheduler driver setup so that the correct IP is published to mesos, namely advertising an IP that matches that of LIBPROCESS_IP (if present)

upstream PRs:
- [ ] https://github.com/kubernetes/kubernetes/pull/21612

there's also some packaging changes required for this:
- mesosphere/kubernetes-mesos DCOS image bootstrap
- multiverse kubernetes package marathon.json changes

jdef:
cherry-picked into v0.7.3, removing cherry-pick label

PR #761, opened 2016-02-01
WIP: v0.7.3 development branch

jdef:

PR #793, opened 2016-02-20
scheduler uber container now expects to be run in docker bridge mode

jdef:
#790

### DIFF
--- a/images/dcos/bootstrap.sh
+++ b/images/dcos/bootstrap.sh
@@ -1,4 +1,14 @@
 #!/bin/sh
+#
+# this bootstrap script is intended to be the startup script for a docker
+# container that runs the k8s-mesos framework scheduler that, at this point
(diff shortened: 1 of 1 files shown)


images/dcos/bootstrap.sh
```diff
@@ -257,7 +268,7 @@ prepare_service ${monitor_dir} ${service_dir} controller-manager ${CONTROLLER_MA
 fdmove -c 2 1
 ${apply_uids}
 /opt/km controller-manager
-  --address=${host_ip}
+  --address=127.0.0.1
```
jdef: may need to use address=0.0.0.0 hostname-override=127.0.0.1 so that we avoid awful hostname problems but still expose the port to the outside world (we'll want to collect metrics at some point)

jdef:
closing this, moving changes to 0.7.3-dev branch #761

PR #800, opened 2016-05-21
Enable authentication against the Mesos cluster.

ivucica:
Introduces two new environment variables:

- MESOS_AUTHENTICATION_PRINCIPAL
- MESOS_AUTHENTICATION_SECRET_FILE

which map to the similarly named flags of the scheduler process.


jdef:
lgtm

jdef:
thanks for the PR!

PR #803, opened 2016-06-30
Fixes for make build

kuba--:
- Fix docker tag in Makefile (removed ':', because of error: 'invalid value for flag --tag:').
- Fix link to glibc-2.21-r2.apk (original one: 'https://circle-artifacts.com/gh/andyshinn/alpine-pkg-glibc/6/artifacts/0/home/ubuntu/alpine-pkg-glibc/packages/x86_64/glibc-2.21-r2.apk' does not work anymore).

### DIFF
--- a/images/dcos/Makefile
+++ b/images/dcos/Makefile
@@ -16,6 +16,9 @@ POSTFIX    ?= $(call LOAD_OR_DEFAULT,POSTFIX,-alpha)
 # NOTE: Do not persist the VERSION. It must be re-computed on every make run.
 GIT_REF_MINOR = $(lastword $(subst -, ,$(GIT_REF)))
 GIT_REF_MAJOR = $(firstword $(subst -, ,$(GIT_REF)))
+ifeq ($(USER),)
+        USER := nobody
+endif
 ifneq ($(wildcard $(BUILD_DIR)),)
 	UNCLEAN_INFIX := -unclean
 endif
--- a/images/dcos/Makefile
+++ b/images/dcos/Makefile
@@ -16,12 +16,14 @@ POSTFIX    ?= $(call LOAD_OR_DEFAULT,POSTFIX,-alpha)
 # NOTE: Do not persist the VERSION. It must be re-computed on every make run.
 GIT_REF_MINOR = $(lastword $(subst -, ,$(GIT_REF)))
 GIT_REF_MAJOR = $(firstword $(subst -, ,$(GIT_REF)))
+ifeq ($(USER),)
+        USER := nobody
+endif
 ifneq ($(wildcard $(BUILD_DIR)),)
 	UNCLEAN_INFIX := -unclean
-	UNCLEAN_PREFIX := $(USER)-
 endif
 ifeq ($(KUBE_ROOT),)
-	VERSION := $(UNCLEAN_PREFIX)$(shell git describe --match $(GIT_REF_MAJOR) --always --tags --dirty)$(UNCLEAN_INFIX)-$(GIT_REF_MINOR)
+	VERSION := $(USER)-$(shell git describe --match $(GIT_REF_MAJOR) --always --tags --dirty)$(UNCLEAN_INFIX)-$(GIT_REF_MINOR)
 else
 	VERSION := $(USER)-$(GIT_REF)-dev
 endif
--- a/images/dcos/Makefile
+++ b/images/dcos/Makefile
@@ -29,7 +29,7 @@ endif
 # compute the Docker image tag
 # NOTE: As VERSION, do not persist DOCKER_IMAGE
 DOCKER_REPO = $(DOCKER_ORG)/kubernetes
-DOCKER_IMAGE = $(DOCKER_REPO):$(VERSION)$(POSTFIX)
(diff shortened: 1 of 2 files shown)


images/dcos/Makefile
```diff
@@ -29,7 +29,7 @@ endif
 # compute the Docker image tag
 # NOTE: As VERSION, do not persist DOCKER_IMAGE
 DOCKER_REPO = $(DOCKER_ORG)/kubernetes
-DOCKER_IMAGE = $(DOCKER_REPO):$(VERSION)$(POSTFIX)
+DOCKER_IMAGE = $(DOCKER_REPO)$(VERSION)$(POSTFIX)
```
jdef: with what version of Docker did the `:` character here stop working?

kuba--: I was building on AMI: dcos-centos7-201606080106 (ami-a891ebc8) 
```
[centos@ip-172-31-5-248 ~]$ sudo docker version
Client:
 Version:      1.11.2
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   b9f10c9
 Built:        Wed Jun  1 21:23:11 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.11.2
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   b9f10c9
 Built:        Wed Jun  1 21:23:11 2016
 OS/Arch:      linux/amd64
```

jdef: Sorry, still don't understand why this failed. According to docker docs the `:tag` notation should still work.
https://docs.docker.com/engine/reference/commandline/tag/

Is there another approach we can take to avoid losing the `:tag` information?

kuba--:
I've updated PR. The problem with tag was because it started with '-' ($USER returned empty string).
Right now build works fine for me (both clean and unclean versions).

images/dcos/Makefile
```diff
@@ -18,10 +18,9 @@ GIT_REF_MINOR = $(lastword $(subst -, ,$(GIT_REF)))
 GIT_REF_MAJOR = $(firstword $(subst -, ,$(GIT_REF)))
 ifneq ($(wildcard $(BUILD_DIR)),)
 	UNCLEAN_INFIX := -unclean
-	UNCLEAN_PREFIX := $(USER)-
```
jdef: could we do something like this instead? `UNCLEAN_PREFIX := $(shell echo $${USER:-nobody})-`

images/dcos/Makefile
```diff
@@ -18,10 +18,9 @@ GIT_REF_MINOR = $(lastword $(subst -, ,$(GIT_REF)))
...
 endif
 ifeq ($(KUBE_ROOT),)
-	VERSION := $(UNCLEAN_PREFIX)$(shell git describe --match $(GIT_REF_MAJOR) --always --tags --dirty)$(UNCLEAN_INFIX)-$(GIT_REF_MINOR)
+	VERSION := $(USER)$(shell git describe --match $(GIT_REF_MAJOR) --always --tags --dirty)$(UNCLEAN_INFIX)-$(GIT_REF_MINOR)
 else
 	VERSION := $(USER)-$(GIT_REF)-dev
```
jdef: seems like the `$(USER)` line here should also be `$(UNCLEAN_PREFIX)` to avoid problems with an empty USER?

kuba--: Check update - I propose to use USER for both clean and unclean, but before that I check if USER is empty:
```
ifeq ($(USER),)
        USER := nobody
endif
```

jdef: cool. does that mean that we can get rid of the changes you made for UNCLEAN_PREFIX?

kuba--:
I've just removed $(UNCLEAN_PREFIX) and replaced it by $USER (to be consistent)

images/dcos/Makefile
```diff
@@ -16,12 +16,14 @@ POSTFIX    ?= $(call LOAD_OR_DEFAULT,POSTFIX,-alpha)
...
+ifeq ($(USER),)
+        USER := nobody
+endif
 ifneq ($(wildcard $(BUILD_DIR)),)
 	UNCLEAN_INFIX := -unclean
-	UNCLEAN_PREFIX := $(USER)-
 endif
 ifeq ($(KUBE_ROOT),)
-	VERSION := $(UNCLEAN_PREFIX)$(shell git describe --match $(GIT_REF_MAJOR) --always --tags --dirty)$(UNCLEAN_INFIX)-$(GIT_REF_MINOR)
+	VERSION := $(USER)$(shell git describe --match $(GIT_REF_MAJOR) --always --tags --dirty)$(UNCLEAN_INFIX)-$(GIT_REF_MINOR)
```
jdef: can you add back the `-` after `$(USER)`?

jdef:
I'm not attached to UNCLEAN_PREFIX but I do want a nicely formatted tag. Just left another comment

kuba--:
sure, it's done.

jdef:
lgtm - please squash and i'll merge

kuba--:
From this machine it will easier if I send another another pull request

kuba--:
squashed commits are on another pr: https://github.com/mesosphere/kubernetes-mesos/pull/805

